PR #10, opened 2017-04-23
allow client to initiate connection with extra headers

alex-shapiro:
Alter `client_async` to let the user initiate the WebSocket handshake with extra headers.


application-developer-DA:
Looks good, thanks!

The only thing to mention is that we'll most likely rewrite this part (see https://github.com/snapview/tungstenite-rs/issues/6).

PR #15, opened 2017-07-26
Only enable tungstenite's "tls" feature when tokio-tungstenite's is enabled

spinda:
Previously tungstenite was always built with its "tls" feature enabled (thus
pulling down native-tls), even when the tls feature wasn't enabled on
tokio-tungstenite. Here we modify Cargo.toml to have tungstenite default to no
active features, then enable the tungstenite/tls feature with
tokio-tungstenite's tls feature. The default configuration remains the same.


application-developer-DA:
Thanks.

PR #36, opened 2018-03-24
Update to tungstenite-rs 0.5.4 so that WebSocketStream Sink implementation can apply back-pressure

### LINKED ISSUE #35: WebSocketStream Sink implementation doesn't apply back-pressure
First, thanks for the sweet crate!

`start_send()` just accepts whatever message you get it and always asks for more! If the messages are coming in faster than Tungstenite can get them out, the underlying `WebSocket::send_queue` just fills up forever.

Besides the memory usage problem, in my server I want to deal with clients that can't take the bandwidth, either by kicking them or by dropping some packets. The wrapper `Sink` I made to accomplish this is especially hacky to work around the fact that `start_send()` will never refuse a message.

Perhaps there should be a configurable max `send_queue` len? I can make a PR if you're into that.

tedsta:
To fix #35

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,9 +6,9 @@ keywords = ["websocket", "io", "web"]
 authors = ["Daniel Abramov <dabramov@snapview.de>", "Alexey Galakhov <agalakhov@snapview.de>"]
 license = "MIT"
 homepage = "https://github.com/snapview/tokio-tungstenite"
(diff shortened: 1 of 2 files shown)


application-developer-DA:
I'm closing this pull request in favor of the new one which will be opened soon to make `tokio-tungstenite` compatible with the future version of `tungstenite-rs`.

PR #37, opened 2018-05-04
Allow constructing WebSocketStream without performing a handshake

bluetech:
Add wrappers around similar constructors for `tungstenite::WebSocket`.

These are useful when doing protocol upgrades in a hyper server, where
the handshake was already performed by the HTTP server.


application-developer-DA:
Thanks.

PR #41, opened 2018-06-19
Expose danger_connect and danger_client_async_tls

mehcode:
Re-implementation of #34 on top of #40

PR #40, opened 2018-06-19
prepare for native-tls 0.2

mehcode:

PR #46, opened 2018-07-19
Update to the newest `tungstenite-rs` version

### LINKED ISSUE #35: WebSocketStream Sink implementation doesn't apply back-pressure

application-developer-DA:
* Fixes #35
* Replaces https://github.com/snapview/tokio-tungstenite/pull/40
* Gives a good background for https://github.com/snapview/tokio-tungstenite/pull/41

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ stream = ["bytes"]
 futures = "0.1.17"
 
 [dependencies.tungstenite]
-version = "0.5.3"
+version = "0.6.0"
 default-features = false
 
 [dependencies.bytes]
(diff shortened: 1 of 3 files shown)


agalakhov:
Please expose the `set_config` API as well.

application-developer-DA:
I also have thought about that and it's trivial to add this feature on the `WebSocket` level in `tokio-tungstenite`, however it would not be much useful in case if the one splits the stream into 2 parts like we show in our examples (client / server). Probably after the migration on new `futures-rs` we have to think how to improve the API so that the dynamic changing of config is also possible.

PR #47, opened 2018-08-09
Update tokio-tls to 0.2 and replace fully tokio-core with tokio

mehcode:
Should be ready to go now @application-developer-DA


application-developer-DA:
Nice! Thanks @mehcode

Issue #44, opened 2018-07-13
0.6 release?


seanmonstar:
Awesome library here! I've been working on a web framework that uses this for websockets, and while getting close to a viable release, it turns out I can't release with a git dependency :D Specifically, being able to use the new tokio crate, and construct a websocket stream after having done my own handshake. Does it seem like 0.6 would be possible soon?

agalakhov:
Yes. It was actually planned for the previous week but this was not done due to other urgent tasks.
Do you have any special wishes for 0.6? Any functionality you're missing?

ParadoxSpiral:
When do you plan to release the new version to crates.io?
I've switched to this crate from `websockets` because this is much nicer, but don't want to push the move until it's not a git dep anymore.

application-developer-DA:
We thought about the release with the current state from master, but there is still one thing we're waiting for: `tokio-tls`. There is a patch in `Cargo.toml` of  `tokio-tungstenite`, without this patch the crate won't compile due to the `tokio-tls` dependency which currently uses the old version of `native-tls`. We hope that `tokio-tls` will get released soon with the new `native-tls` dependency (https://github.com/tokio-rs/tokio-tls/pull/43), at that point we can remove the [temporary] patch from our `Cargo.toml` and release a new version of `tokio-tungstenite`.

zargony:
I'm too using tokio-tungstenite for websocket communication instead of the websocket crate because tokio-tungstenite's is nicer and easier to use.

Speaking of tokio_tls.. What do you think about making the `AutoStream` type public?
That'd be helpful for working with the resulting websocket item of `connect_async`. To handle the resulting websocket (e.g. keep it in a struct or pass it to a function), I need to give it's exact type. However, I can't use its type `WebSocketStream<AutoStream<S>>`, since AutoStream isn't public. So I always end up using the full type `WebSocketStream<StreamSwitcher<TcpStream, TlsStream<TcpStream>>>`, which also means that I need to add tokio_tls as a direct dependency to get the `StreamSwitcher` type, which is a little unconvenient.

agalakhov:
I even feel that `AutoStream` should be a separate crate. There are many cases when the same code could either use TLS or not, and this has nothing to do with WebSocket. We had to reinvent many wheels for Tungstenite.

zargony:
[...]

PR #48, opened 2018-08-09
Add `MaybeTlsStream` wrapper type for streams that might be protected with TLS

zargony:
Connecting to a URL wraps the connection into this wrapper type based on the protocol of the URL (`ws:` or `wss:`). Making this type public allows client applications to specify the websocket's type without importing the tokio_tls crate.

See also #44.

I chose to only export `OptionalTlsStream` if the `tls` feature is active, since its name and purpose doesn't make sense without TLS and may lead to confusion.

### DIFF
--- a/src/connect.rs
+++ b/src/connect.rs
@@ -39,10 +39,12 @@ mod encryption {
     use tungstenite::Error;
     use tungstenite::stream::Mode;
 
-    use stream::NoDelay;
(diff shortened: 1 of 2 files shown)


mehcode:
There is precedence for this to be termed `MaybeTlsSteam`. See https://github.com/hyperium/hyper-tls/blob/master/src/stream.rs

application-developer-DA:
Thanks!

PR #57, opened 2019-04-06
Depend on tungstenite master branch

dfaust:


application-developer-DA:
Oh, we have to think about releasing `tungstenite-rs`, so that we can depend on the version from crates.io

PR #59, opened 2019-06-04
Add AutoPong wrapper

DenisKolodin:
This PR adds a wrapper that automatically handles incoming ping messages. What do you think about it?


application-developer-DA:
I'm not sure if this is required. Replying to ping messages with pongs is a part of the RFC 6455. We automatically add pong reply to the sending queue whenever we receive a ping. It is a part of our core library (`tungstenite-rs`) on which `tokio-tungstenite` relies, see https://github.com/snapview/tungstenite-rs/blob/master/src/protocol/mod.rs#L395

agalakhov:
This is not required. Tungstenite already does this.

DenisKolodin:
Thank you! In this case the PR is really unnecessary :)

PR #62, opened 2019-08-10
Automatically send a close frame when the WebSocketStream gets dropped

najamelan:
This might need some thinking and review. I don't know tokio-tungstenite well
enough to know how this will affect users, but it seems to me that it makes
sense to close the connection without a reason and code when the stream gets
dropped. Otherwise the other end will receive an error that the stream closed
without proper close frame. I feel that dropping is a normal flow, which does not constitute
an error, more so if the error can only be detected remotely.

This avoids having to manually close it always before drop. If this is considered
an error, then I think that it should be caught locally (eg panic in drop if
the connection hasn't been closed) instead of leaving an error to the remote side.

On the implementation. The close function does not seem to be available unless
the proper trait bounds are added. Unfortunately it doesn't seem possible to
create a partial drop impl, so for this to work they have to be added everywhere.

Maybe this is better implemented in tungstenite, but in any case it's annoying
if it's not at least in tokio-tungstenite, because further down the WebSocketStream
might be consumed by stream combinators, meaning that it might no longer be possible
to call close outside of async context. Drop is always outside async context except
if one spawns a future, or calls executor::block_on in order to call close. That seems
not a good alternative to me.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -168,11 +168,11 @@ where
 /// through the respective `Stream` and `Sink`. Check more information about
 /// them in `futures-rs` crate documentation or have a look on the examples
 /// and unit tests for this crate.
-pub struct WebSocketStream<S> {
+pub struct WebSocketStream<S: AsyncRead + AsyncWrite> {
     inner: WebSocket<S>,
 }
 
-impl<S> WebSocketStream<S> {
+impl<S: AsyncRead + AsyncWrite> WebSocketStream<S> {
     /// Convert a raw socket into a WebSocketStream without performing a
     /// handshake.
     pub fn from_raw_socket(
@@ -201,8 +201,16 @@ impl<S> WebSocketStream<S> {
     }
 }
 
+impl<T> Drop for WebSocketStream<T> where T: AsyncRead + AsyncWrite
+{
+    fn drop( &mut self )
+    {
+        let _ = self.inner.close(None);
+    }
+}
+
 #[cfg(feature="stream")]
-impl<S: PeerAddr> PeerAddr for WebSocketStream<S> {
+impl<S: PeerAddr + AsyncRead + AsyncWrite> PeerAddr for WebSocketStream<S> {
     fn peer_addr(&self) -> IoResult<SocketAddr> {
         self.inner.get_ref().peer_addr()
     }


application-developer-DA:
Hm, that's an interesting topic.

Normally we add such changes inside `tungstenite-rs`, but I'm not sure if we have to automatically send a close code when the stream is getting dropped.

First I would like to start with the RFC, according to which the server is the one who initiates a close, i.e. in normal cases the server just sends a close code to the client and then drops the connection after the acknowledgement is received. There are no actions required on the client side in this case (`tungstenite-rs` automatically makes sure that this invariant is honored). There is no other "proper" way to close the websocket connection on a server side. Doing some additional I/O on `Drop` does not seem to be a good idea as it may introduce undesired side effects. If the developer forgets to close the connection on the server side and just drops the connection manually, it's either a mistake done by developer or probably an intention to drop the connection that should consequently cause an error on the other side.

Another problem with doing `Close` on `Drop` means that `Drop` can actually block the thread. In case we're talking about the normal blocking sockets from `std` (something that `tungstenite-rs` does by default), the issue is obvious. In case we're talking about non-blocking sockets, it's not clear what should drive the execution inside `Drop`, i.e. what to do if it returns `WouldBlock`, someone should call `write_pending()` until it's fully written to the stream and then wait for a response (as according to the standard) and if another side does not implement the WebSocket protocol properly (violates the protocol), we can spend some time inside that `Drop`.

The way it's currently written in the pull request won't work as there is no guarantee that the message is sent and/or that the connection will be closed gracefully as there is nothing that drives the execution and polls the stream. Calling `close()` like that may just "schedule" sending of `Close` and drop the object immediately.

IMHO we can't implement it like that in `Drop` and I'm not sure if it's ok to make an assumption that dropping the `WebSocketStream` should cause sending of `Close`.

@agalakhov what do you think?

src/lib.rs
```diff
@@ -201,8 +201,16 @@ impl<S> WebSocketStream<S> {
     }
 }
 
+impl<T> Drop for WebSocketStream<T> where T: AsyncRead + AsyncWrite
+{
+    fn drop( &mut self )
+    {
+        let _ = self.inner.close(None);
```
agalakhov: This won't work properly. It only queues messages for sending and does not guarantee that they will be sent. The network connection will be closed prior to performing close handshake.

agalakhov:
I'm strictly against introducing any side effects in `drop`.
One obvious reason is, dropping may happen as a result of fatal network error. Sending packets in `drop` in this case is impossible.
Another reason is, normal WebSocket connection close requires providing a close code. Closing without a close code cannot be considered normal. 
As of this pull request, calling `close()` in `drop()` just won't work properly since the network connection will be gone immediately without waiting for the handshake to complete.

najamelan:
Ok, so if I get it, the recommended way would be to always require the end user (server application implementor) to explicitly close with a close code before dropping?

application-developer-DA:
Yes, I think sending close on the server side is the expected behavior which normally is written on the server. Also when `tokio-tungstenite` is used, there is a `Sink` trait implementation and you can pass `Close` message to it which will cause the stream to end and eventually will result in dropping of the socket.

> Currently the readme for tungstenite is 1 line, and for tokio-tungstenite. It would be nice to give a bit of an overview in the readme, which could mention that one is supposed to close before drop...

Perhaps you're right. I think the only reason we did not include it into README from the beginning as that we thought that the users who are going to use the crate will be familiar to websockets, I also don't remember mentioning this in other websocket crates (at least back then there were no mentions of that). But yes, we can mention it in our README to avoid confusion.

najamelan:
Ok, thanks for clarifying that all. I'm writing a library which implements AsyncRead/AsyncWrite over websockets, so I will definitely document stuff like this. I don't expect end users to know the RFC at all. I assume people want to be able to send data back and forth to a server. I think they will be grateful if they can do so without scrutinizing 70 pages of technical RFC.

When I refered to the readme, I looked at crates and docs.rs. On github there is a little bit more readme, but you don't really see that if you look at the documentation on docs.rs.

I will scrutinize that part of the RFC, but it's not just reading it once, takes this quote from above:
> Another reason is, normal WebSocket connection close requires providing a close code. Closing without a close code cannot be considered normal. 

The RFC says things like:
```
7.1.5.  The WebSocket Connection Close Code

   As defined in Sections 5.5.1 and 7.4, a Close control frame may
   contain a status code indicating a reason for closure.  A closing of
   the WebSocket connection may be initiated by either endpoint,
   potentially simultaneously. _The WebSocket Connection Close Code_ is
   defined as the status code (Section 7.4) contained in the first Close
   control frame received by the application implementing this protocol.
   If this Close control frame contains no status code, _The WebSocket
   Connection Close Code_ is considered to be 1005.  If _The WebSocket
   Connection is Closed_ and no Close control frame was received by the
   endpoint (such as could occur if the underlying transport connection
   is lost), _The WebSocket Connection Close Code_ is considered to be
   1006.
```
All in all, it's potentially confusing for an end user.

application-developer-DA:
> All in all, it's potentially confusing for an end user, and for example web browser API will allow you to close without specifying a close code.

You can call `close` without specifying a close code, in this case the default close code is considered. There should be a way to say "Ok, now I need to close this connection" on the server. I think sending a close message should be quite simple to use for the end user and IMHO it's simpler than dropping the object in async context.

Also if we refer to [the example](https://github.com/snapview/tokio-tungstenite/blob/master/examples/server.rs) we can see that there is no way to drop the connection inside `for_each`. But there is always a way to send a close message which eventually will cause the `Stream` to return `Ready(None)` and drop the object afterwards.

PR #68, opened 2019-09-09
Conversion to tokio 0.2

dbcfd:
Convert to tokio 0.2


dbcfd:
currently hung up on tokio-dns, looks like the future it returns isn't correct, `Box<Future>` rather than `impl Future` or `Pin<Box<Future>>`

najamelan:
Have you filed an issue at tokio-dns to resolve the blocker?

dbcfd:
Master has the required updates, so using that to keep moving forward.

src/compat.rs
```diff
@@ -0,0 +1,83 @@
+use std::io::{Read, Write};
+use std::pin::Pin;
+use std::task::{Context, Poll};
+
+use tokio_io::{AsyncRead, AsyncWrite};
+
+#[derive(Debug)]
+pub(crate) struct AllowStd<S> {
+    pub(crate) inner: S,
+    pub(crate) context: *mut (),
```
najamelan: Why use void pointers and unsafe when you could use `Option<Context<'_>>`?

Maybe some comments explaining the reasoning behind all this would be good.

src/lib.rs
```diff
@@ -198,138 +238,105 @@ impl<T> Stream for WebSocketStream<T>
...
-
-impl<S: AsyncRead + AsyncWrite> Future for ConnectAsync<S> {
-    type Item = (WebSocketStream<S>, Response);
-    type Error = WsError;
-
-    fn poll(&mut self) -> Poll<Self::Item, WsError> {
-        match self.inner.poll()? {
-            Async::NotReady => Ok(Async::NotReady),
-            Async::Ready((ws, resp)) => Ok(Async::Ready((WebSocketStream::new(ws), resp))),
+    fn poll_close(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
```
najamelan: I just filed #69 which might be relevant to this. I don't think tungstenite returns wouldblock in that case.

src/lib.rs
```diff
@@ -198,138 +238,105 @@ impl<T> Stream for WebSocketStream<T>
...
+        match self.with_context(cx, |s| cvt(s.close(None))) {
+            Ok(()) => Poll::Ready(Ok(())),
+            Err(ref e) if e.kind() == std::io::ErrorKind::WouldBlock => Poll::Pending,
```
najamelan: How does the task get woken up here if we get wouldblock?

src/compat.rs
dbcfd: This is taken from tokio-tls, so currently just trying to adapt rather than adjust since they're also adapting a read/write library there.

src/lib.rs
dbcfd: Not entirely sure on that. As mentioned above, taken from tokio-tls, so would have to ask there.

src/handshake.rs
```diff
@@ -0,0 +1,171 @@
...
+{
+    type Output = Result<Role::FinalResult, Error<Role>>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let mut_self = self.get_mut();
+        let mut s = mut_self.0.take().expect("future polled after completion");
+
+        let machine = s.get_mut();
+
+        (machine.get_mut() as AllowStd<_>).context = cx as *mut _ as *mut ();
```
dbcfd: Currently blocked here. Need to figure out how to get the context back out of the machine to set it appropriately.

src/stream.rs
```diff
@@ -3,102 +3,57 @@
...
-        match *self {
-            Stream::Plain(ref mut s) => s.set_nodelay(nodelay),
-            Stream::Tls(ref mut s) => s.set_nodelay(nodelay),
+impl<S: AsyncRead, T: AsyncRead> AsyncRead for Stream<S, T> {
+    fn poll_read(
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        buf: &mut [u8],
+    ) -> Poll<std::io::Result<usize>> {
+        match self.project() {
```
dbcfd: This is a project type, so may have to just match deref rather than pin project, but that requires creating a pin afterwards. map_unchecked_mut might be better?

dbcfd:
> Just asking some questions, I haven't fully reviewed everything.
> 
> I do feel this lacks integration testing, verifying every possible scenario, but then again, tokio-tungstenite didn't have those before either.

Will look at add at least a couple more once I get something that seems like it might work.

src/compat.rs
najamelan: Ok, I see. I also just realized that I confused with storing a Waker, which I usually do in order to wake up a task when some event happens. That is easier than storing `Context<'_>` because of the lifetime. Still think that ideally we avoid using unsafe for things like this.

src/lib.rs
najamelan: Ok, yeah, I think this is wrong then. If you return Pending without an inner future returning Pending, you have to wake up the task somewhere. I don't know if there is a recommended way of doing that, but I usually store the Waker in an option and when other code in my object detects the right condition, it wakes up that task. In this case I doubt (haven't dug into it completely and it's not documented :( ) that tungstenite returns wouldblock here at all...

dbcfd: Will look at tungstenite some more. If it never returns wouldblock, then we can nuke compat, and use tokio blocking around the read/write.

dbcfd: Yeah, looks like any would blocks are converted into other types, which we can chain futures off of.

najamelan: I had the impression it won't return wouldblock for `close`. In principle it might for other things, but since we are working on AsyncRead/AsyncWrite, and we are have the impl of `std::Read`/`std::Write`, we could intercept the Pendings there and tungstenite would not have to return wouldblock.

One exception would be the `MessageQueueFull` error probably.

CryZe:
This doesn't build without the tls feature.

dbcfd:
@CryZe This was actually an issue in tungstenite and is fixed in https://github.com/snapview/tungstenite-rs/pull/73. cargo update and this branch should build now.

CryZe:
Nice, another issue I ran into was that the pin-project `0.4.0-alpha.10` pulled in here is causing a cargo update failure when the new hyper is also in the dependency graph. But I don't think it's reasonable to ask for you to downgrade to their version. That's a general problem with all these alpha versions that are all semver incompatible.

dbcfd:
Matching hyper for now, since one of the reasons for this work was to use it in warp.

CryZe:
If I'm sending two messages before trying to receive any I get:
```
thread 'main' panicked at 'assertion failed: !self.context.is_null()', C:\Users\Christopher Serr\.cargo\git\checkouts\tokio-tungstenite-42c59291ea04b35c\bea935e\src\compat.rs:44:13 
```

dbcfd:
@CryZe Added a test which reproduces this behavior, attempting to determine cause and fix it.

dbcfd:
The multiple write failure is due to tungstenite write_message calling flush. This means that Sink cannot be used for send, since it does not include a context.

CryZe:
That seems to have fixed the bug for me :)

CryZe:
Is there a reason this doesn't use Sink for sending elements?

dbcfd:
@CryZe Sink cannot be used because tungstenite write_message is also calling flush. Sink write does not provide a context, so no context is present when write_message then calls flush.

One possibility is to separate the the write and flush into separate functions in tungstenite (that write_message uses), so then we could implement Sink in tokio-tungstenite.

dbrgn:
Async/await is now available on Beta, so we don't even need nightly anymore! :tada: 

https://blog.rust-lang.org/2019/09/30/Async-await-hits-beta.html

CryZe:
Yeah looks like tokio and futures have released versions that are compatible with the beta, so I'd appreciate if this PR would be updated to those versions as well.

skaufhold:
Just wanted to say I'm pretty excited about this PR, I've been trying the most recent version out for a side project the last few days. It's just very smooth integration, I'd say currently this is shaping up to become the go-to for anyone building a websocket based application using async/await.

The only issue I've noticed so far is due to not implementing `Sink` in this version, as mentioned above - it means you can't split the stream into Sender and Receiver because `StreamExt::split` requires `Self: Sink`, which can be a killer for some applications

anlumo:
Can someone point me to a way to read and write to the same Websocket with this implementation without `Sink`? Both `poll_next` and `send` need a `&mut self`, so I can't do both without splitting.

agalakhov:
It is doable via `.into_future()` of the `Stream` trait for reading. The whole is likely to be called from `unfold()` or `loop_fn()`.

anlumo:
I just discovered that it does work when I remove my call to `map` on the stream. However, I need that to be able to `select` on two different streams with different item types. The idea is that I have an `futures::channel::mpsc` as the incoming message channel from other parts of the program, so these other parts can communicate with the websocket.

Seems like this is one of the applications where the `Sink` issue is a killer, as @skaufhold mentioned.

dbcfd:
@skaufhold I will look about adding a split method to this, so that Sink is not required.

@anlumo It will look something like this

```
while let Some(try_msg) = websocket_stream.next().await {
    let msg = try_msg.expect("Failed to receive message");
    channel.send(msg);
}```

You can also do something like ```
while let Some(try_msg) = websocket_stream.next().await {
  let msg = try_msg.expect("Failed to receive message");
  let response = transform(msg);
  websocket_stream.send(response).await.expect("Failed to send response");
}```

The examples (especially autobahn) show this off better https://github.com/snapview/tokio-tungstenite/pull/68/files#diff-9da84c2001004ed41d861e7b1fc0850fR37

**Comment by mcseemk** [comment]
> I just discovered that it does work when I remove my call to `map` on the stream. However, I need that to be able to `select` on two different streams with different item types. The idea is that I have an `futures::channel::mpsc` as the incoming message channel from other parts of the program, so these other parts can communicate with the websocket.
> 
> I also can't use `forward`, since that requires a `Sink` as well.
> 
> Seems like this is one of the applications where the `Sink` issue is a killer, as @skaufhold mentioned.

I use the following workaround:

```rust
let (mut ws_stream, _) = match connect_async(...).await {
    Ok(x) => x,
    Err(e) => {}
};

let mut heartbeat = Interval::new(Duration::from_secs(3));

'inner: loop {
    let either1 = select(receive_channel.next(), select(heartbeat.next(), ws_stream.next())).await;
    match either1 {
        Either::Left((msg, _)) => {
            // received message via channel, process it
        }
        Either::Right((either2, _)) => {
            match either2 {
                Either::Left(_) => {
                    // heartbeat, do whatever
                }
                Either::Right((msg, _)) => {
                    // websocket message received, process it
                }
            }
        }
    }
}

**Comment by dbcfd** [comment]
@mcseemk You can also use the select macro from futures

```
let opt_value = select! {
    opt_hb = heartbeat.next() => opt_hb.map(Received::Heartbeat),
    opt_recv = receive_channel.next() => opt_recv.map(Received::Outgoing),
    opt_msg = ws_stream.next() => opt_msg.map(Received::Incoming),
};
if let Some(v) = opt_value {
    match opt_value {
        Received::Outgoing(msg) => ws_stream.send(msg).await.expect("Failed to send response"),
```

**Comment by mcseemk** [comment]
BTW, I use this WIP PR in production, works great.
Would be great if it was merged sooner than later.

**Comment by anlumo** [comment]
Using `future::select` with `.next()` instead of `stream::select` works, thank you!

**Comment by application-developer-DA** [comment]
> BTW, I use this WIP PR in production, works great.
> Would be great if it was merged sooner than later.

It seems like async is getting stable. When the API is more or less stable, I'll have a closer look at this PR (the final version), maybe cleaning up the implementation a bit (there are still some places which I'm a bit concerned of).

**Comment by dbcfd** [comment]
Rebased, removed wip.

A couple quick passes at trying to do split, keep hitting a point where `write_message` calling flush is problematic. Would either have to change tungstenite, or live with no split.

**Comment by skaufhold** [comment]
The select! macro seems to require `FusedFuture`/`FusedStream` to be implemented to call `is_terminated`, which is currently missing. The select function already works great though.

**Comment by carllerche** [comment]
Did you figure out DNS? Tokio 0.2 includes DNS APIs.

**Comment by dbcfd** [comment]
@carllerche was using tokio-dns-unofficial, did not realize it got added to tokio. Will switch over.

**Comment by carllerche** [comment]
@dbcfd :+1: Feel free to ping me w/ any questions.

**Comment by Drevoed** [comment]
Can't we store the context (and the buffer) in the layer of async websocket implementation? So that sink can be implemented or sync tungstenite doesn't offer the flush API?

**Comment by dbcfd** [comment]
@Drevoed Sink write comes in without a context, so there's no context to be stored.

If we really want sink capability, probably better just to make a version of tungstenite that is pure async.

**Comment by najamelan** [comment]
@dbcfd I don't feel like I completely understand all factors here, but from what I can figure, flush is mainly meaningful if you buffer something in your type, which as far as I can tell is not the case here. This makes me think that it's probably fine to return `Ok(())`  if tungstenite decides to call flush when there is no context. 

When the user calls `poll_flush` on the `Sink`, there is a context, and there is intend of flushing, so now we can forward the call to `poll_flush` of the `AsyncWrite`.

**Comment by dbcfd** [comment]
@najamelan Your suggestion is similar to the approach taken by jxs in his warp upgrade branch: https://github.com/jxs/warp/blob/master/src/filters/ws.rs#L269

I will make that change, but definitely agree it needs more testing around it

**Comment by cere42** [comment]
I think there's  something weird going on with the dependencies in these latest versions. When specified as a dependency in another project, this version doesn't compile with an `unresolved import tokio_net::tcp` error.

However when cloned and compiled directly it all compiles fine. It has something to do with the dev dependencies, since when I comment out `tokio = "0.2.0-alpha.6"` I can provoke the error even in the cloned project. I don't understand why though, since I'm pretty sure `tokio_net::tcp` should come from the tokio-net dependency, not from the tokio dependency. This happens for me since the version where tokio-dns-unofficial was removed.

**Comment by cere42** [comment]
Found the issue! The tokio-net entry is missing a `features = ["tcp"]` line.

**Comment by dbcfd** [comment]
Thanks @cere42

**Comment by GoWebProd** [comment]
Can't run your example:

```
error[E0277]: the trait bound `dyn futures::future::Future<Error = tungstenite::error::Error, Item = (tokio_tungstenite::WebSocketStream<tokio_tungstenite::stream::Stream<tokio_tcp::stream::TcpStream, tokio_tls::TlsStream<tokio_tcp::stream::TcpStream>>>, tungstenite::handshake::client::Response)> + std::marker::Send: std::marker::Unpin` is not satisfied
  --> src/main.rs:57:30
   |
57 |     let (mut ws_stream, _) = connect_async(url).await.expect("Failed to connect");
   |                              ^^^^^^^^^^^^^^^^^^^^^^^^ the trait `std::marker::Unpin` is not implemented for `dyn futures::future::Future<Error = tungstenite::error::Error, Item = (tokio_tungstenite::WebSocketStream<tokio_tungstenite::stream::Stream<tokio_tcp::stream::TcpStream, tokio_tls::TlsStream<tokio_tcp::stream::TcpStream>>>, tungstenite::handshake::client::Response)> + std::marker::Send`
   |
   = note: required because of the requirements on the impl of `core::future::future::Future` for `std::boxed::Box<dyn futures::future::Future<Error = tungstenite::error::Error, Item = (tokio_tungstenite::WebSocketStream<tokio_tungstenite::stream::Stream<tokio_tcp::stream::TcpStream, tokio_tls::TlsStream<tokio_tcp::stream::TcpStream>>>, tungstenite::handshake::client::Response)> + std::marker::Send>`
   = note: required by `std::future::poll_with_tls_context`

error[E0277]: the trait bound `dyn futures::future::Future<Error = tungstenite::error::Error, Item = (tokio_tungstenite::WebSocketStream<tokio_tungstenite::stream::Stream<tokio_tcp::stream::TcpStream, tokio_tls::TlsStream<tokio_tcp::stream::TcpStream>>>, tungstenite::handshake::client::Response)> + std::marker::Send: core::future::future::Future` is not satisfied
  --> src/main.rs:57:30
   |
57 |     let (mut ws_stream, _) = connect_async(url).await.expect("Failed to connect");
   |                              ^^^^^^^^^^^^^^^^^^^^^^^^ the trait `core::future::future::Future` is not implemented for `dyn futures::future::Future<Error = tungstenite::error::Error, Item = (tokio_tungstenite::WebSocketStream<tokio_tungstenite::stream::Stream<tokio_tcp::stream::TcpStream, tokio_tls::TlsStream<tokio_tcp::stream::TcpStream>>>, tungstenite::handshake::client::Response)> + std::marker::Send`
   |
   = note: required because of the requirements on the impl of `core::future::future::Future` for `std::boxed::Box<dyn futures::future::Future<Error = tungstenite::error::Error, Item = (tokio_tungstenite::WebSocketStream<tokio_tungstenite::stream::Stream<tokio_tcp::stream::TcpStream, tokio_tls::TlsStream<tokio_tcp::stream::TcpStream>>>, tungstenite::handshake::client::Response)> + std::marker::Send>`
   = note: required by `std::future::poll_with_tls_context`

GoWebProd:
Oh, sorry, i'm trying to dep master branch :)

bstrie:
@dbcfd Can I ask what the status of the PR is? Are there big open questions, or is it just waiting for review?

dbcfd:
@bstrie I think it's just waiting for review now. Could probably use more tests, but between what is there, the testing I've done, and the testing I think others have done, it should be good to go.

dbrgn:
I'm working on a project that uses this branch, and haven't run into issues so far. I haven't done a lot of in-depth testing or load testing though.

Maybe after a review this code could be released as a beta release, for more real-life feedback?

mcseemk:
I've been using it in production 24/7 for about three weeks already, no issues so far.

bstrie:
> It seems like async is getting stable. When the API is more or less stable, I'll have a closer look at this PR (the final version), maybe cleaning up the implementation a bit (there are still some places which I'm a bit concerned of).

@application-developer-DA , does "the API" here refer to Tokio? I'm under the impression that the 0.2 alphas are indeed more-or-less stable API-wise. Given that we have some experience with using this PR in the wild, would you consider merging this and making an alpha release to crates.io? I agree that making a stable release should wait for all dependencies to have stable releases, but the way that the async ecosystem is proceeding at the moment is that all the foundational libraries are making alpha releases so that their dependents can themselves prepare for stable async/await (and so on and so on up the dependency chain). To give an example, a published alpha of this library would allow Warp to publish an alpha with async websocket support.

dbrgn:
Something I noticed when playing around with this branch: When the call to `tokio_tungstenite::accept_async(stream)` fails, the stream has been consumed and cannot be used in the error handling code.

I would like to write a message to the stream before closing, if the handshake failed. Should that be addressed in this branch, or should I open a new issue?

application-developer-DA:
> does "the API" here refer to Tokio? I'm under the impression that the tokio 0.2 alphas are indeed more-or-less stable API-wise.

Yes, I was referring to the tokio ecosystem. My biggest concern so far is the `AllowStd` (it looks like a bunch of hacks, which I believe are [for now] required, until it all gets stable and futures from `std` are compatible with tokio(?)) and unsafe code together with raw pointers. I'm not sure if we can get rid of them right now though.

> would  you consider accepting this and making an alpha release to crates.io? I  agree that making a stable release should wait for all dependencies to  have stable releases, but the way that the async ecosystem is proceeding  at the moment is that all the foundational libraries are making alpha  releases so that their dependents can themselves prepare for stable  async/await (and so on and so on up the dependency chain). To give an  example, a published alpha of this library would allow Warp to publish  an alpha with async websocket support (which would then allow anyone  depending on Warp to prepare for async/await, and so on).

@agalakhov , what do you think?

application-developer-DA:
> Something I noticed when playing around with this branch: When the call to tokio_tungstenite::accept_async(stream) fails, the stream has been consumed and cannot be used in the error handling code.
> I would like to write a message to the stream before closing, if the handshake failed. Should that be addressed in this branch, or should I open a new issue?

If I got you right, we already have an enhancement issue for that in `tungstenite-rs`: https://github.com/snapview/tungstenite-rs/issues/51

dbrgn:
@application-developer-DA ah, thanks, you're right! Feel free to collapse my comments.

sdroege:
FYI I've also did a port to [`async-std`](https://github.com/async-rs/async-std) here: https://github.com/sdroege/async-tungstenite

The examples and tests are not all ported yet, but apart from that it's working fine.

sdroege:
> I think it's just waiting for review now. Could probably use more tests, but between what is there, the testing I've done, and the testing I think others have done, it should be good to go.

@dbcfd The autobahn-client / autobahn-server tests are failing on Travis it seems. For the server it's lots of threads panicking, for the client it's:

```
Running test case ID 1.1.1 for agent Tungstenite from peer tcp4:127.0.0.1:57538

thread 'main' panicked at 'Failed to get message: Io(Os { code: 104, kind: ConnectionReset, message: "Connection reset by peer" })', src/libcore/result.rs:1165:5

note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

That probably needs some debugging.

dbcfd:
@sdroege It seems like that is ok (tests still pass). It seems to be a fuzz test in travis against autobahn.

sdroege:
> @sdroege It seems like that is ok (tests still pass). It seems to be a fuzz test in travis against autobahn.

Ok then :) I didn't look closer at that test, only saw it failing on travis. I didn't run it locally yet.

Cargo.toml
```diff
@@ -8,21 +8,25 @@ license = "MIT"
...
-connect = ["tokio-dns-unofficial", "tokio-tcp", "stream"]
+connect = ["tokio-net", "stream"]
 tls = ["tokio-tls", "native-tls", "stream", "tungstenite/tls"]
 stream = ["bytes"]
 
 [dependencies]
-futures = "0.1.23"
-tokio-io = "0.1.7"
+log = "0.4"
+futures = { version = "0.3", features = ["async-await"] }
```
glebpom: This will break things until tokio is using futures-preview

src/connect.rs
```diff
@@ -85,7 +53,7 @@ mod encryption {
 pub use self::encryption::MaybeTlsStream;
 
 #[cfg(not(feature = "tls"))]
-mod encryption {
+pub(crate) mod encryption {
     use futures::{future, Future};
```
glebpom: This imports are not used

Cargo.toml
cere42: Related tokio issue: https://github.com/tokio-rs/tokio/issues/1708 There should be a new alpha release soon with the new stable futures as dependencies.

Weirdly, I'm currently using this updated pull request in a project together with the current tokio alpha (the futures-preview based one) and it seems to work.

frol: Tokio master branch is already using it, so it is just a matter of a new release. Also, this change did not break things for me :man_shrugging:

zargony: The `async-await` feature is enabled by default in futures 0.3, so no need to explicitly enable it anymore.

glebpom: I think it would make sense to release the alpha version of the crate, as `hyper` and `reqwest` do. In this case, it's better to use fixed versions of `futures-preview` and `tokio` for now.

frol:
I have bumped into a deadlock in client implementation and a panic in server implementation when I concurrently write and read the data non-stop. To reproduce, run the `server` example (`cargo run --example server --release`) and the following echo-client:

```rust
//! A simple example of hooking up stdin/stdout to a WebSocket stream.
//!
//! This example will connect to a server specified in the argument list and
//! then bombard it with messages while still reading all the replies.
//!
//! You can use this example together with the `server` example.

use std::env;

use futures::{SinkExt, StreamExt};
use log::*;
use tungstenite::protocol::Message;

use tokio_tungstenite::connect_async;

#[tokio::main]
async fn main() {
    let _ = env_logger::try_init();

    // Specify the server address to which the client will be connecting.
    let connect_addr = env::args()
        .nth(1)
        .unwrap_or_else(|| panic!("this program requires at least one argument"));

    let url = url::Url::parse(&connect_addr).unwrap();

    // After the TCP connection has been established, we set up our client to
    // start forwarding data.
    //
    // First we do a WebSocket handshake on a TCP stream, i.e. do the upgrade
    // request.
    //
    // Next, we start bombarding the server with messages and reading the replies.
    let (ws_stream, _) = connect_async(url).await.expect("Failed to connect");
    let (mut ws_tx, mut ws_rx) = ws_stream.split();
    info!("WebSocket handshake has been successfully completed");

    tokio::spawn(async move {
        // Discard all the incomming messages
        while let Some(_) = ws_rx.next().await {}
    });

    loop {
        ws_tx
            .send(Message::text("ping".to_owned()))
            .await
            .expect("Failed to send request");
    }
}
```

After just a few seconds of a max load, it gets stuck. If the client does not dead lock, the server will panic in 5-10 seconds:

```
thread 'tokio-runtime-worker-4' panicked at 'Failed to send response: Io(Kind(WouldBlock))', src/libcore/result.rs:1165:5
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/libunwind.rs:88
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src/libstd/sys_common/backtrace.rs:77
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
             at src/libstd/sys_common/backtrace.rs:61
   4: core::fmt::write
             at src/libcore/fmt/mod.rs:1028
   5: std::io::Write::write_fmt
             at src/libstd/io/mod.rs:1412
   6: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:65
   7: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:50
   8: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:188
   9: std::panicking::default_hook
             at src/libstd/panicking.rs:205
  10: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:464
  11: std::panicking::continue_panic_fmt
             at src/libstd/panicking.rs:373
  12: rust_begin_unwind
             at src/libstd/panicking.rs:302
  13: core::panicking::panic_fmt
             at src/libcore/panicking.rs:139
  14: core::result::unwrap_failed
             at src/libcore/result.rs:1165
  15: <std::future::GenFuture<T> as core::future::future::Future>::poll
  16: <std::panic::AssertUnwindSafe<F> as core::ops::function::FnOnce<()>>::call_once
  17: std::panicking::try::do_call
  18: __rust_maybe_catch_panic
             at src/libpanic_unwind/lib.rs:78
  19: tokio_executor::threadpool::task::Task::run
  20: tokio_executor::threadpool::worker::Worker::run_task
  21: tokio_executor::threadpool::worker::Worker::run
  22: tracing_core::dispatcher::with_default
  23: std::thread::local::LocalKey<T>::with
  24: tokio::runtime::threadpool::builder::Builder::build::{{closure}}
  25: std::thread::local::LocalKey<T>::with
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

I am yet to learn what is going on.

Just for the record, I have also [tried tokio master branch](https://github.com/frol/tokio-tungstenite/commit/d59b60211aaba7ce490eaed6191ad65862f4bef9) (they have done quite a major [project restructure](https://github.com/tokio-rs/tokio/pull/1766)). The issue persists.

Can someone reproduce this?

My environment:

* Arch Linux x86-64 (Linux kernel 4.19.84-1-lts)
* Rust 1.39.0 stable and 1.40.0-nightly (1423bec54 2019-11-05)

frol:
There is also a memory leak somewhere on the server (I have replaced the unbounded channels with the bounded ones and I still see +50MB RAM per second under the echo-load).

Just for the record, here is the server implementation using bounded channels.

```rust
//! A chat server that broadcasts a message to all connections.
//!
//! This is a simple line-based server which accepts WebSocket connections,
//! reads lines from those connections, and broadcasts the lines to all other
//! connected clients.
//!
//! You can test this out by running:
//!
//!     cargo run --example server 127.0.0.1:12345
//!
//! And then in another window run:
//!
//!     cargo run --example client ws://127.0.0.1:12345/
//!
//! You can run the second command in multiple windows and then chat between the
//! two, seeing the messages from the other client as they're received. For all
//! connected clients they'll all join the same room and see everyone else's
//! messages.

use std::env;
use std::io::Error;

use futures::channel::mpsc::{Receiver, Sender};
use futures::{SinkExt, StreamExt};
use log::*;
use std::net::{SocketAddr, ToSocketAddrs};
use tokio::net::{TcpListener, TcpStream};
use tungstenite::protocol::Message;

struct Connection {
    addr: SocketAddr,
    rx: Receiver<Message>,
    tx: Sender<Message>,
}

async fn handle_connection(connection: Connection) {
    let mut connection = connection;
    let mut received = 0;
    while let Some(msg) = connection.rx.next().await {
        received += 1;
        if received % 1000 == 0 {
            println!("R: {}", received);
        }
        info!("Received a message from {}: {}", connection.addr, msg);
        connection
            .tx
            .send(msg)
            .await
            .expect("Failed to forward message");
    }
}

async fn accept_connection(stream: TcpStream) {
    let addr = stream
        .peer_addr()
        .expect("connected streams should have a peer address");
    info!("Peer address: {}", addr);

    let mut ws_stream = tokio_tungstenite::accept_async(stream)
        .await
        .expect("Error during the websocket handshake occurred");

    info!("New WebSocket connection: {}", addr);

    // Create a channel for our stream, which other sockets will use to
    // send us messages. Then register our address with the stream to send
    // data to us.
    let (mut msg_tx, msg_rx) = futures::channel::mpsc::channel(100);
    let (response_tx, mut response_rx) = futures::channel::mpsc::channel(100);
    let c = Connection {
        addr: addr,
        rx: msg_rx,
        tx: response_tx,
    };
    tokio::spawn(handle_connection(c));

    let mut sent = 0;
    while let Some(message) = ws_stream.next().await {
        let message = message.expect("Failed to get request");
        msg_tx
            .send(message)
            .await
            .expect("Failed to forward request");
        if let Some(resp) = response_rx.next().await {
            ws_stream.send(resp).await.ok();
            sent += 1;
            if sent % 1000 == 0 {
                println!("S: {}", sent);
            }
        }
    }
}
#[tokio::main]
async fn main() -> Result<(), Error> {
    let _ = env_logger::try_init();
    let addr = env::args().nth(1).unwrap_or("127.0.0.1:8080".to_string());
    let addr = addr
        .to_socket_addrs()
        .expect("Not a valid address")
        .next()
        .expect("Not a socket address");

    // Create the event loop and TCP listener we'll accept connections on.
    let try_socket = TcpListener::bind(&addr).await;
    let socket = try_socket.expect("Failed to bind");
    let mut incoming = socket.incoming();
    info!("Listening on: {}", addr);

    while let Some(stream) = incoming.next().await {
        let stream = stream.expect("Failed to accept stream");
        tokio::spawn(accept_connection(stream));
    }

    Ok(())
}
```

dbcfd:
This PR is still blocked on https://github.com/snapview/tungstenite-rs/commit/2b2e66326453709d3b3df50c1a18e7be96f2221f being merged as well. I've updated the dependencies to work with stable

Cargo.toml
dbcfd: We're still waiting on a release of tungstenite-rs before an alpha of this can be released.

src/connect.rs
dbcfd: Fixed now. Forgot to try building without default features.

Cargo.toml
application-developer-DA: > We're still waiting on a release of tungstenite-rs before an alpha of this can be released.

`tungstenite-rs` 0.9.2 has been published today.

FreeFull:
When I attempt to use tokio-tungstenite with the latest git version of tokio, I get the following runtime error:
https://gist.github.com/FreeFull/936f68d4d7410b5b89c1e4552fb6c318

When I try to use the non-git version of tokio (0.2.0-alpha.6) then it seems like the Stream trait implemented by tokio-tungstenite isn't the same as the Stream trait from tokio, due to tokio using futures-preview.

jbg:
@FreeFull yes, tokio hasn't released an alpha that uses the new non-preview `futures` release yet.

dbcfd:
> When I attempt to use tokio-tungstenite with the latest git version of tokio, I get the following runtime error:
> https://gist.github.com/FreeFull/936f68d4d7410b5b89c1e4552fb6c318
> 
> When I try to use the non-git version of tokio (0.2.0-alpha.6) then it seems like the Stream trait implemented by tokio-tungstenite isn't the same as the Stream trait from tokio, due to tokio using futures-preview.

Can you grab a backtrace?

FreeFull:
Here is the backtrace:
```
    Finished dev [unoptimized + debuginfo] target(s) in 0.07s
     Running `target/debug/tungstenite-test`
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Io(Custom { kind: Other, error: "no current reactor" })', src/libcore/result.rs:1165:5
stack backtrace:
   0:     0x55bb6261bc54 - backtrace::backtrace::libunwind::trace::hda41dbcdfba36aa0
                               at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.37/src/backtrace/libunwind.rs:88
   1:     0x55bb6261bc54 - backtrace::backtrace::trace_unsynchronized::h1a8d6e1f8cb3f5d4
                               at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.37/src/backtrace/mod.rs:66
   2:     0x55bb6261bc54 - std::sys_common::backtrace::_print_fmt::h610c4127487e10da
                               at src/libstd/sys_common/backtrace.rs:76
   3:     0x55bb6261bc54 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h0722dc552e01bd1d
                               at src/libstd/sys_common/backtrace.rs:60
   4:     0x55bb62637a5c - core::fmt::write::h01edf6dd68a42c9c
                               at src/libcore/fmt/mod.rs:1030
   5:     0x55bb62618987 - std::io::Write::write_fmt::hf15985f193f03c04
                               at src/libstd/io/mod.rs:1412
   6:     0x55bb6261e105 - std::sys_common::backtrace::_print::hd8d5d08a1795e743
                               at src/libstd/sys_common/backtrace.rs:64
   7:     0x55bb6261e105 - std::sys_common::backtrace::print::hf89a79e3921a2366
                               at src/libstd/sys_common/backtrace.rs:49
   8:     0x55bb6261e105 - std::panicking::default_hook::{{closure}}::h3a8f42beb3bb8ae3
                               at src/libstd/panicking.rs:196
   9:     0x55bb6261ddf6 - std::panicking::default_hook::h8f803b0bc31a5c37
                               at src/libstd/panicking.rs:210
  10:     0x55bb6261e805 - std::panicking::rust_panic_with_hook::h825f041245da8739
                               at src/libstd/panicking.rs:473
  11:     0x55bb6261e3a2 - std::panicking::continue_panic_fmt::hbe0378e33481e81b
                               at src/libstd/panicking.rs:380
  12:     0x55bb6261e296 - rust_begin_unwind
                               at src/libstd/panicking.rs:307
  13:     0x55bb62634f6a - core::panicking::panic_fmt::h527855ce0bc891f6
                               at src/libcore/panicking.rs:85
  14:     0x55bb62635067 - core::result::unwrap_failed::ha8b77e6004f0ba38
                               at src/libcore/result.rs:1165
  15:     0x55bb623ea8c6 - core::result::Result<T,E>::unwrap::h9c4a18bddfd47892
                               at /rustc/4560ea788cb760f0a34127156c78e2552949f734/src/libcore/result.rs:933
  16:     0x55bb6239f6b8 - tungstenite_test::main::{{closure}}::hc520572dd907aabf
                               at src/main.rs:9
  17:     0x55bb623d857b - <std::future::GenFuture<T> as core::future::future::Future>::poll::{{closure}}::h594f63d1036b5517
                               at /rustc/4560ea788cb760f0a34127156c78e2552949f734/src/libstd/future.rs:43
  18:     0x55bb623d66e9 - std::future::set_task_context::ha253a2a679540505
                               at /rustc/4560ea788cb760f0a34127156c78e2552949f734/src/libstd/future.rs:79
  19:     0x55bb623d7e06 - <std::future::GenFuture<T> as core::future::future::Future>::poll::h960367452532ffc2
                               at /rustc/4560ea788cb760f0a34127156c78e2552949f734/src/libstd/future.rs:43
  20:     0x55bb623af234 - tokio::runtime::enter::Enter::block_on::h44799b034fa8c0b9
                               at /home/freefull/.cargo/git/checkouts/tokio-377c595163f99a10/320c84a/tokio/src/runtime/enter.rs:99
  21:     0x55bb623e7c83 - tokio::runtime::thread_pool::ThreadPool::block_on::{{closure}}::he8d84410248e7ab4
                               at /home/freefull/.cargo/git/checkouts/tokio-377c595163f99a10/320c84a/tokio/src/runtime/thread_pool/mod.rs:122
  22:     0x55bb623e7ea7 - tokio::runtime::global::with_state::{{closure}}::h2cc077bc10112728
                               at /home/freefull/.cargo/git/checkouts/tokio-377c595163f99a10/320c84a/tokio/src/runtime/global.rs:107
  23:     0x55bb623cb9d5 - std::thread::local::LocalKey<T>::try_with::ha4bd50c4cf7ad150
                               at /rustc/4560ea788cb760f0a34127156c78e2552949f734/src/libstd/thread/local.rs:262
  24:     0x55bb623c8baf - std::thread::local::LocalKey<T>::with::he4e04b50437e218d
                               at /rustc/4560ea788cb760f0a34127156c78e2552949f734/src/libstd/thread/local.rs:239
  25:     0x55bb623e7d9e - tokio::runtime::global::with_state::hd5a038d8e7d2f280
                               at /home/freefull/.cargo/git/checkouts/tokio-377c595163f99a10/320c84a/tokio/src/runtime/global.rs:90
  26:     0x55bb623e80bb - tokio::runtime::global::with_thread_pool::h953dc5b390fc23d9
                               at /home/freefull/.cargo/git/checkouts/tokio-377c595163f99a10/320c84a/tokio/src/runtime/global.rs:83
  27:     0x55bb623e7be0 - tokio::runtime::thread_pool::ThreadPool::block_on::hc38ab7435e8c038b
                               at /home/freefull/.cargo/git/checkouts/tokio-377c595163f99a10/320c84a/tokio/src/runtime/thread_pool/mod.rs:120
  28:     0x55bb623e931b - tokio::runtime::Runtime::block_on::{{closure}}::h41f98f13b7f99c75
                               at /home/freefull/.cargo/git/checkouts/tokio-377c595163f99a10/320c84a/tokio/src/runtime/mod.rs:334
  29:     0x55bb623c4ede - tokio::time::clock::variant::Clock::enter::hc272be0f1818745e
                               at /home/freefull/.cargo/git/checkouts/tokio-377c595163f99a10/320c84a/tokio/src/time/clock.rs:38
  30:     0x55bb623d9188 - tokio::runtime::time::variant::with_default::h4d84c4439c84f2b2
                               at /home/freefull/.cargo/git/checkouts/tokio-377c595163f99a10/320c84a/tokio/src/runtime/time.rs:34
  31:     0x55bb623e8efd - tokio::runtime::handle::Handle::enter::{{closure}}::hbbc5945393bcc7f0
                               at /home/freefull/.cargo/git/checkouts/tokio-377c595163f99a10/320c84a/tokio/src/runtime/handle.rs:93
  32:     0x55bb623e90d8 - tokio::blocking::pool::Spawner::enter::{{closure}}::h7fe2c6f23e324fab
                               at /home/freefull/.cargo/git/checkouts/tokio-377c595163f99a10/320c84a/tokio/src/blocking/pool.rs:159
  33:     0x55bb623ca8f5 - std::thread::local::LocalKey<T>::try_with::h6e4d31ece80a6c43
                               at /rustc/4560ea788cb760f0a34127156c78e2552949f734/src/libstd/thread/local.rs:262
  34:     0x55bb623c84ff - std::thread::local::LocalKey<T>::with::h6799c53821ff351c
                               at /rustc/4560ea788cb760f0a34127156c78e2552949f734/src/libstd/thread/local.rs:239
  35:     0x55bb623e8fc8 - tokio::blocking::pool::Spawner::enter::h3c19b5ad3f664479
                               at /home/freefull/.cargo/git/checkouts/tokio-377c595163f99a10/320c84a/tokio/src/blocking/pool.rs:144
  36:     0x55bb623e8e26 - tokio::runtime::handle::Handle::enter::h5955af0c425bb33b
                               at /home/freefull/.cargo/git/checkouts/tokio-377c595163f99a10/320c84a/tokio/src/runtime/handle.rs:90
  37:     0x55bb623e91c0 - tokio::runtime::Runtime::block_on::he040f534724390aa
                               at /home/freefull/.cargo/git/checkouts/tokio-377c595163f99a10/320c84a/tokio/src/runtime/mod.rs:329
  38:     0x55bb623cf388 - tungstenite_test::main::h14b3d35b96de697b
                               at src/main.rs:6
  39:     0x55bb623a9550 - std::rt::lang_start::{{closure}}::hfceacd101fad637a
                               at /rustc/4560ea788cb760f0a34127156c78e2552949f734/src/libstd/rt.rs:64
  40:     0x55bb6261e233 - std::rt::lang_start_internal::{{closure}}::ha04574f12d97cbe2
                               at src/libstd/rt.rs:49
  41:     0x55bb6261e233 - std::panicking::try::do_call::h7c2a8488f72db90c
                               at src/libstd/panicking.rs:292
  42:     0x55bb62621c9a - __rust_maybe_catch_panic
                               at src/libpanic_unwind/lib.rs:80
  43:     0x55bb6261edbd - std::panicking::try::hc3a9b5da4250385d
                               at src/libstd/panicking.rs:271
  44:     0x55bb6261edbd - std::panic::catch_unwind::hf27600bf8c37809a
                               at src/libstd/panic.rs:394
  45:     0x55bb6261edbd - std::rt::lang_start_internal::h409d4f2fe51133b0
                               at src/libstd/rt.rs:48
  46:     0x55bb623a9529 - std::rt::lang_start::h96afeeb45e712753
                               at /rustc/4560ea788cb760f0a34127156c78e2552949f734/src/libstd/rt.rs:64
  47:     0x55bb623cf3da - main
  48:     0x7f7263bee153 - __libc_start_main
  49:     0x55bb6239d17e - _start
  50:                0x0 - <unknown>
```

dbcfd:
@FreeFull can you switch from tokio master to version "0.2.0-alpha.6"

FreeFull:
With version "0.2.0-alpha.6", and the suitable code adjustments, the `.split()` method does not work (there are two incompatible Stream and Sink traits, one from tokio (which uses futures-preview) and one from futures 0.3). It used to work before the bb0957b28415381827b11b38f31343b183eacbd7 commit in your branch.

dbcfd:
@FreeFull good to hear. I keep hoping that today is the day that at least a new tokio alpha gets released, so glad that this works with a few gotchas on alpha.6

Atul9:
Hi @dbcfd I have run `cargo fmt` on this branch and opened up a PR at https://github.com/dbcfd/tokio-tungstenite/pull/4, please check

frol:
Can someone review the issues I mentioned above? Here is the detailed reproduction: https://github.com/frol/tungstenite-async-bug-reproduction

I have implemented 4 servers and 3 clients, and send and receive 10 million echo messages:

| Client \ Server | actix               | async-std | tokio 0.2-alpha6  | tokio master |
| --------------- | ------------------- | --------- | ----------------- | ------------ |
| async-std       | 25s                 | 32s       | 65s               | 45s          |
| tokio 0.2alpha6 | 42s                 | 59s       | stuck             | 44s          |
| tokio master    | 18s <sup>(*1)</sup> | stuck     | stuck             | stuck        |

(*1) The implementation stuck if you increase the number of iterations
as it sends messages faster than receives.

(stuck) client stops sending or receiving messages for no apparent reason.

dbcfd:
@frol Tokio master uses the new scheduler, which did have a bug recently https://github.com/tokio-rs/tokio/commit/7c8b8877d440629ab9a27a2c9dcef859835d3536

frol:
@dbcfd I am using https://github.com/tokio-rs/tokio/commit/69975fb9601bbb21659db283d888470733bae660 which includes the referenced commit in both my tokio-tungstenite fork and the reproduction example, so there must be something else.

carllerche:
@frol Could you help me repro (maybe jump in the #tokio-dev discord).

I ran your example, and what I see on the server is a panic:

```
thread 'tokio-runtime-worker' panicked at 'Failed to send response: Io(Kind(WouldBlock))', src/libcore/result.rs:1165:5
```

On the client, I see `Received: DONE`. Is this expected?

carllerche:
Edit, in the client, I also sometimes get `ERR: Io(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" }`.

frol:
@carllerche Where can I find a link to the Discord server?

> On the client, I see Received: DONE. Is this expected?

Meh, no. That happens since `ws_rx.next().await` returned None and the loop just ended.

`thread 'tokio-runtime-worker' panicked at 'Failed to send response: Io(Kind(WouldBlock))', src/libcore/result.rs:1165:5`

Well, this is an expected panic on `ws_stream.send(resp).await.expect("Failed to send response");`, it is just a result of the fact that the network queue is full, which happens due to the fact that the client did not keep up with reading from the queue. The panic on the server leads to `ERR: Io(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" }` on the client, so these two are just the result of not reading messages in the first place.

hawkw:
When running the repro with the tokio master client and the Actix server, it looks like all the worker threads are parked polling the reactor (per lldb). Maybe wakeups are being lost?
```
(lldb) thread list
Process 21962 stopped
  thread #1: tid = 0xda51b, 0x00007fff761db86a libsystem_kernel.dylib`__psynch_cvwait + 10, queue = 'com.apple.main-thread', stop reason = signal SIGSTOP
* thread #2: tid = 0xda51e, 0x00007fff761de78e libsystem_kernel.dylib`kevent + 10, name = 'tokio-runtime-worker'
  thread #3: tid = 0xda51f, 0x00007fff761de78e libsystem_kernel.dylib`kevent + 10, name = 'tokio-runtime-worker'
  thread #4: tid = 0xda520, 0x00007fff761de78e libsystem_kernel.dylib`kevent + 10, name = 'tokio-runtime-worker'
  thread #5: tid = 0xda521, 0x00007fff761de78e libsystem_kernel.dylib`kevent + 10, name = 'tokio-runtime-worker'
```

carllerche:
@frol https://discord.gg/u2Jspb

frol:
FYI, it seems that all the people who tried to reproduce the issue mentioned above confirmed the observed behavior. Thus, it is not safe to merge this PR.

dbcfd:
@frol This seems to not be an issue with this pr so much as tokio. At some points, WouldBlock is returned, and things are polled again. However eventually WouldBlock is returned, but no polls occur. Given that this same code works with async-std, I think there's a difference in how the executor is working.

sdroege:
> This seems to not be an issue with this pr so much as tokio.

I agree with this interpretation and it would be good to get this merged so that further work can happen on top of it. The tokio issue will be fixed sooner or later :)

src/lib.rs
```diff
@@ -160,176 +174,240 @@ where
...
-/// Future returned from accept_async() which will resolve
-/// once the connection handshake has finished.
-pub struct AcceptAsync<S: AsyncRead + AsyncWrite, C: Callback> {
-    inner: MidHandshake<ServerHandshake<S, C>>,
-}
+    fn start_send(mut self: Pin<&mut Self>, item: Message) -> Result<(), Self::Error> {
+        match (*self).with_context(None, |s| s.write_message(item)) {
+            Ok(()) => Ok(()),
+            Err(::tungstenite::Error::Io(ref err))
+                if err.kind() == std::io::ErrorKind::WouldBlock =>
```
carllerche: I think this is the bug. When the inner stream returns Pending / WouldBlock, it needs to register the async context in order to notify the task. Doing this here will result in lost notifications and deadlock.

carllerche:
I think I found the source of the bug and highlighted it in the code. The writes hang and when the readiness event comes from tokio, the tasks don't seem to resume. I believe it is due to the line I highlighted.

src/lib.rs
dbcfd: Send method on Futures::Sink doesn't provide a context, and splitting the stream requires a Sink implementation.

Are we just getting lucky with async-std, or is it doing something behind the scene to handle this?

dbcfd: Actually, looking at the Sink code, we should get a registration 
```rust
    if let Some(item) = this.item.take() {
            let mut sink = Pin::new(&mut this.sink);
            match sink.as_mut().poll_ready(cx)? {
                Poll::Ready(()) => sink.as_mut().start_send(item)?,
                Poll::Pending => {
                    this.item = Some(item);
                    return Poll::Pending;
                }
            }
        }

        // we're done sending the item, but want to block on flushing the
        // sink
        ready!(Pin::new(&mut this.sink).poll_flush(cx))?;```

`poll_flush` will call `write_pending` in the sync code, which I think might be blocking the pool due to WouldBlock. Trying to see if there is a way to identify WouldBlock before calling `write_pending`.

Comment (hawkw) [reply src/lib.rs]: > Are we just getting lucky with async-std, or is it doing something behind the scene to handle this?

I'm fairly sure that we're just getting lucky due to some detail of `async-std`'s scheduler, or something. The contract of Rust's async model is that when we yield, we have to notify the task via the context. I can't imagine `async-std` not also assuming that contract?

Comment (carllerche) [reply src/lib.rs]: When start_send is called, it must simply queue the data, not attempt to use the socket. The actual write to the socket should happen in poll_flush. 

One possible fix would be to always return would block from AllowStd without trying to use the AsyncWrite if no context is set. Then the call to poll_flush will go down the path again and trigger the actual write with the context. 
M

Comment (sdroege) [reply src/lib.rs]: > One possible fix would be to always return would block from AllowStd without trying to use the AsyncWrite if no context is set. Then the call to poll_flush will go down the path again and trigger the actual write with the context.

That's actually what is happening already from what I can see [here](https://github.com/snapview/tokio-tungstenite/pull/68/files/a27bd616b46fb0c8aab94b913f1c565b32f8e5ae#diff-eeffc045326f81d4c46c22f225d3df90R49) before calling any on the actual `AsyncWrite`. That `Pending` is then converted to `WouldBlock` immediately [here](https://github.com/snapview/tokio-tungstenite/pull/68/files/a27bd616b46fb0c8aab94b913f1c565b32f8e5ae#diff-eeffc045326f81d4c46c22f225d3df90R101). Which one layer up is then converted to an `Ok(())` in `start_send()` [here](https://github.com/snapview/tokio-tungstenite/pull/68/files/a27bd616b46fb0c8aab94b913f1c565b32f8e5ae#diff-b4aea3e418ccdb71239b96952d9cddb6R316). So then `poll_flush()` will call into `write_pending()`, which then does actual writing, would most likely register for being woken up again later.

What am I missing?

Comment (frol) [reply src/lib.rs]: > Are we just getting lucky with async-std, or is it doing something behind the scene to handle this?

I believe, it is just a bit of pure luck. Tokio 0.2-alpha.6 and async-std don't deadlock since their speed of sending matches the speed of receiving messages, so `WouldBlock` just never occurs. I am thankful to @carllerche for help!

Yet, I still wonder why Tokio master has this disbalance of sent and received messages.

Comment (carllerche) [reply src/lib.rs]: You are correct, this is not the bug. I should have checked the actual implementation instead of stopping at "something looks fishy". I dug in more and I have a new hypothesis. I will comment below as it is not related to this thread.

**Comment by carllerche** [comment]
I did more digging and I noticed Tokio is firing the notification correctly but the wrong task (the reader) is being notified. I dug into tungstenite and found this:

https://github.com/snapview/tungstenite-rs/blob/master/src/protocol/mod.rs#L256

The reader task is writing... and that switches the task to be woken...

File: src/lib.rs (hunk 8 above)
Comment (sdroege) [reply src/lib.rs]: Thanks for looking :)

**Comment by sdroege** [comment]
> The reader task is writing... and that switches the task to be woken...

Hmm interesting.

You mean inside tokio only a single task is remembered to be woken up for the write side of a TCP stream? Wouldn't that also fall apart if e.g. simply using the write side of a TCP stream inside an `Arc<futures::lock::Mutex<_>>` from two different tasks (which is basically what `StreamExt::split()` does in the end).

From the tokio-tungstenite side I don't see how the wrong context would be stored in `AllowStd` temporarily: both reader and writer side would require a mutable reference to call anything, so there can only be one operation at a time. And each operation would set the current context temporarily and ensure that it's unset again once the operation returns.

**Comment by carllerche** [comment]
split works because TcpStream tracks 2 tasks: one for reading, one for writing. This is having problems because multiple tasks are "waiting" for the same op, which violates the contract (Future, Sink, Stream, ... all limit a single task per handle

**Comment by sdroege** [comment]
> split works because TcpStream tracks 2 tasks: one for reading, one for writing. This is having problems because multiple tasks are "waiting" for the same op, which violates the contract (Future, Sink, Stream, ... all limit a single task per handle

Interesting, I didn't consider that. So having e.g. a `Sink` inside a `Arc<Mutex<_>>` would also fail the same way and be considered wrong? OOC, where is that specified in the docs? It makes sense for performance reasons, but maybe replacing the task with a different one could panic instead of silently doing the wrong thing?

**Comment by carllerche** [comment]
Replacing the task with a new one is not wrong. It is necessary to support moving values to different tasks. I would suggest opening an issue on std / futures to request added docs.

**Comment by sdroege** [comment]
Good point. I'll create an issue in a bit. Do you happen to know how this same problem is handled elsewhere? E.g. I would expect the same (a read requiring a write) to potentially happen with the async adapters for the TLS libraries.

One potential solution would seem to put the whole thing behind a `futures::lock::Mutex` so that only one task at a time can read or write and never both at once. Any better ideas? :)

**Comment by sdroege** [comment]
This probably fixes it https://github.com/dbcfd/tokio-tungstenite/pull/5

**Comment by carllerche** [comment]
Historically, the pattern used in this case is to use a dedicated "driver" task. ie do all read / writes from a single task.

I don't think the mutex strategy above would work because the same handle needs to be kept until it is ready.

Another strategy would be to have `AllowStd` to maintain separate read / write wakers and do a proxy strategy (similar to how FuturesUnordered does it).

**Comment by sdroege** [comment]
> I don't think the mutex strategy above would work because the same handle needs to be kept until it is ready.

Indeed, that makes no sense after rethinking it. A driver task would make the whole thing considerably more complicated, that does not seem very appealing either :)

**Comment by carllerche** [comment]
@sdroege the third option would work then, a "fan out" waker. `AllowStd` would maintain "waker proxies" so if both the reader / writer tasks end up writing and both are blocked on readiness, the waker proxy notifies both (fan out).

**Comment by sdroege** [comment]
> @sdroege the third option would work then, a "fan out" waker.

That seems like a good and relatively non-intrusive approach. I'll try reserving some time in the next days to take a look at implementing that if nobody else is faster.

Do you know of an existing, known-to-work implementation of that somewhere that can be taken as a starting point?

**Comment by hawkw** [comment]
@sdroege I'm planning on looking into writing a reusable implementation of something like that, probably to be released in `tokio-util`. I'll let you know when that's ready to try out.

**Comment by sdroege** [comment]
I came up with https://github.com/dbcfd/tokio-tungstenite/pull/6 but didn't test yet if this actually solves the problem, or if it makes any sense at all. Needs some thinking but it's late here :)

**Comment by FreeFull** [comment]
The stable version of tokio 0.2.0 is now out

**Comment by sdroege** [comment]
> I came up with [dbcfd#6](https://github.com/dbcfd/tokio-tungstenite/pull/6) but didn't test yet if this actually solves the problem

Seems like it solves the problem, I can't reproduce it anymore at least with the testcase given above. Now time to clean up my code, I guess.

**Comment by sdroege** [comment]
> > I came up with [dbcfd#6](https://github.com/dbcfd/tokio-tungstenite/pull/6) but didn't test yet if this actually solves the problem
> 
> Seems like it solves the problem, I can't reproduce it anymore at least with the testcase given above. Now time to clean up my code, I guess.

I did a new release of async-tungstenite with these fixes. tokio-tungstenite is blocked by the wider ecosystem being partially stuck with tokio 0.2.0-alpha currently (specifically [hyper](https://github.com/dbcfd/tokio-tungstenite/pull/6#pullrequestreview-324816519))

**Comment by jbg** [comment]
@sdroege tokio-tungstenite doesn't seem to have a dependency on hyper. can you clarify why marging this or making a release is blocked on hyper? (btw, I imagine a lot of crates are taking a bit longer to follow-up from the tokio stable release, with these days being Thanksgiving in the US.)

**Comment by sdroege** [comment]
On 30 November 2019 16:08:43 EET, jbg <notifications@github.com> wrote:
>@sdroege tokio-tungstenite doesn't seem to have a dependency on hyper.
>can you clarify why marging this or making a release is blocked on
>hyper?

I don't know, you'll have to ask there :) I'm just giving an update here from a comment in the PR.

My guess is that Danny has crates that use both tokio-tungstenite and hyper, and doesn't want to break the dependencies there currently and because of that waits until everything is at tokio 0.2.
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

**Comment by jbg** [comment]
that's what we have versioning for! someone has to make a release first :)

**Comment by dbcfd** [comment]
@jbg I can't do releases on this repo, but here's some refs for the different tokio versions.

0.10.0-alpha.2 targeting tokio 0.2.0-alpha.6: https://github.com/dbcfd/tokio-tungstenite/releases/tag/v0.10.0-alpha.2
0.10.0 targeting tokio 0.2: https://github.com/dbcfd/tokio-tungstenite/releases/tag/v0.10.

**Comment by nemasu** [comment]
Hello @dbcfd , I'm fiddling with the server and client examples.
The description says it broadcasts to all connected clients, but it seems to only echo messages back to the client that sent it.

**Comment by nemasu** [comment]
Hello @dbcfd , I made a [PR](https://github.com/dbcfd/tokio-tungstenite/pull/8) to update the server example. Seems like you need to hit return in the client terminal(s) before any new messages show up, I'm pretty sure that's the clients fault though.

**Comment by CryZe** [comment]
What's the status of this?

**Comment by bstrie** [comment]
I believe that the acceptance of this PR was blocked on the next stable hyper release? That's just happened ( https://seanmonstar.com/post/189594157852/hyper-v013 ), so are there any remaining issues here?

**Comment by dbcfd** [comment]
@bstrie Nope :D

**Comment by dbcfd** [comment]
@application-developer-DA ready for review/merge

File: Cargo.toml [hunk 9]
```diff
@@ -8,45 +8,34 @@ license = "MIT"
 homepage = "https://github.com/snapview/tokio-tungstenite"
 documentation = "https://docs.rs/tokio-tungstenite/0.9.0"
 repository = "https://github.com/snapview/tokio-tungstenite"
-version = "0.9.0"
+version = "0.10.0-alpha.1"
```
bstrie: Given that this version of the PR is no longer depending on any alpha libraries, do we want to drop the "alpha" here and release this as `0.10.0`?

Cargo.toml
```diff
@@ -8,45 +8,34 @@ license = "MIT"
 homepage = "https://github.com/snapview/tokio-tungstenite"
 documentation = "https://docs.rs/tokio-tungstenite/0.9.0"
```
samrg472: Documentation URL should point to 0.10.0 to match the correct version.

Cargo.toml
```diff
@@ -21,7 +21,7 @@ stream = []
 log = "0.4"
 futures = "0.3"
 pin-project = "0.4"
-tokio = "0.2"
+tokio = { version = "0.2", default-features = false, features = ["io-util"] }
```
bstrie: Nit, but tokio's Cargo.toml has:
```
# Include nothing by default
default = []
```
so `default-features = false` is a no-op.

CryZe: Well that's assuming they won't add some. But I guess that's somewhat unlikely.

vorot93:
@application-developer-DA Any chance this could be merged ASAP?

kellytk:
What, if anything, could be done to progress this?

bstrie:
I would suspect that the maintainers are reasonably predisposed during the holidays, although I suppose it wouldn't hurt to ping @agalakhov as well.

agalakhov:
I'll check this during the holidays.

vorot93:
@application-developer-DA while examples are definitely important, I strongly suggest merging this as is if it works. There is too much downstream blocked from upgrading to Tokio 0.2 because this PR has not been merged yet.

agalakhov:
Checked this partially. This pull request requires more cleanup to be merged. I'm working on it.

vorot93:
@agalakhov If it's in the working state it could be possible to merge as is and keep iterating in subsequent commits. At this point *some* working release is much better than no release at all!

agalakhov:
I have to check for some potential security and stability issues first. Especially handshake code has to be carefully checked. Now it is much better than initially (no more `unsafe`), but I'm still suspicious about some places. I just don't want to see any CVE filed.

agalakhov:
Ok, I checked this. While `handshake` submodule could be improved, I did not found any security related problems yet. It seems to be safe to merge. Thank you.

PR #71, opened 2020-01-07
 Use tokio::net::ToSocketAddrs instead of sync version from std

sdroege:
And fix a couple of clippy warnings while we're at it.


agalakhov:
Thank you!

PR #79, opened 2020-01-25
Update server.rs

morozik11:


sdroege:
How does it not work for you?

agalakhov:
If you believe you found a bug, please file a bug report.
If you believe you fixed a bug, please provide a fix.
There is no reason to edit the code by just adding comments that "it does not work".
Rejected.

Issue #82, opened 2020-02-02
Real word example question


hbgl:
I want to adapt the https://github.com/snapview/tokio-tungstenite/blob/46dfd9ed3ee75b0261e9f5f71c8e70492407248b/examples/autobahn-server.rs to my use case. In addition to echoing every message the server should send a message to the client every second.

The new `handle_connection` function would look like this:

```
async fn handle_connection(peer: SocketAddr, stream: TcpStream) -> Result<()> {
    let mut ws_stream = accept_async(stream).await.expect("Failed to accept");

    info!("New WebSocket connection: {}", peer);

    let mut msg_fut = ws_stream.next();
    loop {
        let delay_fut = delay_for(Duration::from_millis(1000));
        
        match select(msg_fut, delay_fut).await {
            Either::Left((msg, _)) => {
                match msg {
                    Some(msg) => {
                        let msg = msg?;
                        if msg.is_text() || msg.is_binary() {
                            ws_stream.send(msg).await?;
                        }
                        msg_fut = ws_stream.next();
                    },
                    None => {
                        break;
                    },
                };
            },
            Either::Right((_, msg_fut_reuse)) => {
                msg_fut = msg_fut_reuse;
                // Next line error message:
                //
                // error[E0499]: cannot borrow `ws_stream` as mutable more than once at a time
                // --> src\main.rs:48:17
                //    |
                // 25 |     let mut msg_fut = ws_stream.next();
                //    |                       --------- first mutable borrow occurs here
                // ...
                // 29 |         match select(msg_fut, delay_fut).await {
                //    |                      ------- first borrow later used here
                // ...
                // 48 |                 ws_stream.send(Message::Text("hello".to_owned())).await?;
                //    |                 ^^^^^^^^^ second mutable borrow occurs here
                ws_stream.send(Message::Text("hello".to_owned())).await?;
            },
        }
    }

    Ok(())
}
```

Sadly this does not compile. I have tried to come to an agreement with the borrow checker but I cannot figure out how to reuse `msg_fut_reuse` for the next iteration of the loop.

Can anyone help me with this? I would gladly contribute it as an example later.

agalakhov:
You borrowed the stream for receiving. It is impossible to send at the same time to the same object. Use the `split()` method (trait `StreamExt`) to split the object into two parts:
```
let (sender, receiver) = ws_stream.split();
... receiver.next(); ...
... sender.send(...); ...
```

PR #84, opened 2020-02-02
Add example of server periodically updating client

hbgl:
Example from issue #82 

This example shows how to both handle WebSocket messages and send updates to the client periodically.

A real world example would be a dashboard application that lets the user send commands in a request/response model but also updates the dashboard's stats periodically.


agalakhov:
Thank you!

PR #85, opened 2020-02-04
Fix clippy warning about ok_or() followed by a function call

sdroege:
warning: use of `ok_or` followed by a function call
       --> src/connect.rs:132:10
        |
    132 |         .ok_or(Error::Url("Url scheme not supported".into()))?;
        |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try this: `ok_or_else(|| Error::Url("Url scheme not supported".into()))`


application-developer-DA:
Thanks.

PR #86, opened 2020-02-13
build(deps): minimize unnecessary dependencies by using `futures-{channel,util}` directly instead of `futures`

ErichDonGubler:


ErichDonGubler:
@application-developer-DA I don't believe you could get the same dependency graph (minus a top-level `futures`, obviously) by configuring the features of the `futures` crate -- which is why I set out to implement this this way originally. Perhaps I missed something?

PR #90, opened 2020-03-05
Don't project if you require Unpin.

najamelan:
This is the counter proposition of #89. Since it's rather convoluted to remove the `Unpin` requirement on the underlying stream users can wrap in `WebSocketStream` and I don't know of any use cases for it, I propose to keep the `Unpin` and consequently remove pin-projection from tokio-tungstenite. This also sheds a dependency.

The first commit is completely self contained. Nothing outside of this file will observe any difference. I think it's good to apply it no matter what except if choosing to remove Unpin bounds everywhere as discussed in #89. 

The second is a bit more complicated. It's concerned with the handshake, which takes place before there even is a `WebSocketStream`. There is a type `Role::InternalStream`, which comes from tungstenite, which uses trait bounds `Read + Write` and doesn't mention `Unpin`. The current state allowed this type to be `!Unpin`, although I suspect that normally this is the same underlying connection that is used for wrapping in `WebSocketStream`, so there probably already wasn't any way for users to pass a `!Unpin` type in here. 

However, I haven't actually tested this assumption. If the assumption holds, I suppose this is not a breaking change. If it changes what people could pass in, it is a breaking change.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,6 @@ stream = []
 [dependencies]
 log = "0.4"
 futures-util = { version = "0.3", default-features = false, features = ["async-await", "sink", "std"] }
-pin-project = "0.4"
 tokio = { version = "0.2", default-features = false, features = ["io-util"] }
 
 [dependencies.tungstenite]
--- a/src/handshake.rs
+++ b/src/handshake.rs
@@ -1,7 +1,6 @@
 use crate::compat::{AllowStd, SetWaker};
 use crate::WebSocketStream;
 use log::*;
-use pin_project::pin_project;
 use std::future::Future;
 use std::io::{Read, Write};
 use std::pin::Pin;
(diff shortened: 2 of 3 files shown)


najamelan:
@sdroege what do you think of this solution? Would you like an equivalent PR on async-tungstenite?

sdroege:
I think it makes sense, yes

najamelan:
oh, and what I didn't think of before, users should still be able use `!Unpin` streams, they just have to pin them first, so that's not to bad. I havn't tested that, but AsyncRead is implemented for Pinned AsyncRead in any case.

najamelan:
Note that async-tungstenite, which is a fork with almost identical code chose this path. I think #89 only makes sense if a real path is demonstrated which allows using !Unpin streams, which I didn't find easily.

application-developer-DA:
> Thanks for tackling this issue, @najamelan . I did not think that deep about this issue previously. Well, for me it seems that the approach suggested here (#90) would indeed be a better solution for now and additionally we get rid of the pin-project dependency "for free".

> @agalakhov what do you think? #89 vs #90 ?

@agalakhov , what's your opinion on that?

agalakhov:
I think #90. It looks like safer approach to me.

application-developer-DA:
Superseded by https://github.com/snapview/tokio-tungstenite/pull/194

Issue #97, opened 2020-03-30
more descriptive error messages for connect_async


EverlastingBugstopper:
When sending a custom `http::Request` to `connect_async` I get an error `Io(Custom { kind: Other, error: "" })` and I'm not sure what went wrong or how to address it.

example code

```rust
use tokio_tungstenite::connect_async;
use http::Request;

let socket_url = "ws://example.com/socket"
let socket_request = Request::builder()
  .uri(socket_url)
  .header("X-Auth-Token", token_value)
  .body(())
  .unwrap();
match connect_async(socket_request).await {
  Ok((ws_stream, _)) => {
    println!("handle stream!");
  },
  Err(e) => println!("{:?}", e)
}
```

application-developer-DA:
Hm, interesting. I think something similar has already been raised recently, but back then the error did not originate from our crate.

EverlastingBugstopper:
Yeah - there is no additional output with `RUST_LOG=debug`, so can't help there.

EverlastingBugstopper:
I pushed up an example [here](https://github.com/EverlastingBugstopper/tungstenite-example) if you'd like to play around, should be able to just `cargo run` :)

Dushistov:
I suppose I also met this problem.
My server replied with: "HTTP/1.1 403 Forbidden\r\n\r\n",
if there are no specific http header. So if I run any websocket example
without providing this header I would get 403.

[tungstenite client example](https://github.com/snapview/tungstenite-rs/blob/master/examples/client.rs) output is:
'Can't connect: Http(403)'

So it can recognize http error in server response,
but  [tokio-tungstenite client example](https://github.com/snapview/tokio-tungstenite/blob/master/examples/client.rs) output is:
'Failed to connect: Io(Custom { kind: Other, error: "" })'

Why the error mapping mechanizm doesn't work the same way as in tungstenite-rs ?

sdroege:
Hmm, with `tokio-tungstenite` the examples above give indeed that useless error. With `async-tungstenite` (which is basically the same code base but refactored to work with tokio, async-std and others instead of just tokio) I get `Io(Custom { kind: Other, error: "HTTP error: 404 Not Found" })`

Dushistov:
> I get Io(Custom { kind: Other, error: "HTTP error: 404 Not Found" })

It is better, but still not as good as `tungstenite::Error::Http(404)`,
because of the first one you  can only write to log, while the second one
[...]

PR #103, opened 2020-05-22
Forward handshake errors directly instead of wrapping them in a Other…

sdroege:
… IO error

This allows getting HTTP status codes back to the caller, for example,
and allows proper handling of handshake errors.

Fixes https://github.com/snapview/tokio-tungstenite/issues/97


Dushistov:
@application-developer-DA

Any hope for merging this PR?

Issue #111, opened 2020-07-09
Skip certificate verification


elbaro:
```
thread 'main' panicked at 'Failed to connect: Tls(Ssl(Error { code: ErrorCode(1), cause: Some(Ssl(ErrorStack([Error { code: 337047686, library: "SSL routines", function: "tls_process_server_certificate", reason: "certificate verify failed", file: "ssl/statem/statem_clnt.c", line: 1913 }]))) }, X509VerifyResult { code: 19, error: "self signed certificate in certificate chain" }))', src/main.rs:91:33
```

I found some incomplete PRs, for example this one: https://github.com/snapview/tokio-tungstenite/pull/41.
Can I skip this verification?

application-developer-DA:
Fixed by #116

PR #116, opened 2020-08-05
Add the option to configure the TLS connection.

gabi-250:
This PR addresses #111. 

Instead of exposing a `danger_client_async_tls` function, I've decided to add the option to pass a `TlsConnector` to `client_async_tls_with_config` (which will enable people to add root certificates to the connector, or disable certificate validation altogether).

I'm happy to some up with a different solution if you disagree with this change.


application-developer-DA:
Thanks!

PR #122, opened 2020-10-13
Migrate to GitHub Actions

nickelc:
The PR adds an optimized GitHub Actions workflow and splits the original build job into two jobs, one for the basic tests and another for the Autobahn Testsuite.

see https://github.com/nickelc/tokio-tungstenite/actions/runs/304339616

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,73 @@
+name: CI
+
+on: [push, pull_request]
+
+jobs:
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        rust:
+          - stable
+          - beta
+          - nightly
+
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
+
+      - name: Install toolchain
+        uses: hecrj/setup-rust-action@v1
+        with:
+          rust-version: ${{ matrix.rust }}
+
+      - name: Install dependencies
+        run: sudo apt-get install libssl-dev
+
+      - name: Check default-features
+        run: cargo check
+
+      - name: Check tls
+        run: cargo check --features tls
+
+      - name: Test
+        run: cargo test --release
+
+  autobahn:
+    name: Autobahn tests
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        rust:
+          - stable
+          - beta
+          - nightly
+
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
+
+      - name: Install toolchain
+        uses: hecrj/setup-rust-action@v1
+        with:
+          rust-version: ${{ matrix.rust }}
+
+      - name: Install dependencies
+        run: |
+          sudo apt-get install libssl-dev python-unittest2
+          sudo pip install urllib3[secure] autobahntestsuite
+
+      - name: Build Autobahn TestSuite for client
+        run: cargo build --release --example autobahn-client
+
+      - name: Build Autobahn TestSuite for server
+        run: cargo build --release --example autobahn-server
+
+      - name: Running Autobahn TestSuite for client
+        run: ./scripts/autobahn-client.sh
+
+      - name: Running Autobahn TestSuite for server
+        run: ./scripts/autobahn-server.sh


application-developer-DA:
Thanks for the PR.

I'm not gonna lie, I'm not very familiar with Github Actions. Does it have any benefits over Travis? I'm just wondering, because it seems like Github Actions version is a bit bigger and complicated.

nickelc:
Better UI integration without leaving the website and workflows can run on forks without creating PRs.
It's also faster with the parallel execution of `cargo test` and the Autobahn testsuite.

i think it's not really that complicated for a GHA workflow but I'm biased because of the amount of workflows i've already written.
I could drop the 2nd commit and keep the travis build, so you can get a feel for it.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,78 @@
...
+
+      - name: Install toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: ${{ matrix.rust }}
+          profile: minimal
+          override: true
+
+      - name: Install dependencies
+        run: |
```
application-developer-DA, line 29: Is `|` a typo here?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,78 @@
...
+          sudo apt-get install libssl-dev
+
+      - name: Check default-features
+        run: cargo check
+
+      - name: Check tls
+        run: cargo check --features tls
+
+      - name: Test
+        run: cargo test
```
application-developer-DA, line 39: Was it meant to be `cargo test --release`?

.github/workflows/ci.yml
nickelc: it's the yaml syntax for keep newlines https://yaml-multiline.info/

.github/workflows/ci.yml
nickelc: oh yeah, it's with --release in the travis config. It's originated from the workflow I created for async-tungstenite.

nickelc, line 39:
```suggestion
        run: cargo test --release
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,78 @@
...
+      - name: Install toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: ${{ matrix.rust }}
+          profile: minimal
+          override: true
+
+      - name: Install dependencies
+        run: |
+          sudo apt-get install libssl-dev
```
nickelc, line 30:
```suggestion
        run: sudo apt-get install libssl-dev
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,78 @@
...
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        rust:
+          - stable
+          - beta
+          - nightly
```
application-developer-DA, line 15: I'm not sure about these. Generally it's good that our project will get built on `beta` and `nightly`, but what if it does not compile on `nightly` or `beta` for some reason and the whole PR is marked as wrong because of that? @agalakhov what do you think?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,78 @@
...
+  autobahn:
+    name: Autobahn tests
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        rust:
+          - stable
+          - beta
+          - nightly
```
application-developer-DA, line 49: I'm not familiar with GitHub Actions, but maybe you can advise: is there a possibility to define rust matrix once, so that we don't have repeat the common code for different "jobs" (or "targets").

application-developer-DA:
> i think it's not really that complicated for a GHA workflow but I'm biased because of the amount of workflows i've already written.

Indeed, most of the parts are not complicated, except for probably `actions/checkout@v2` and `actions-rs/toolchain@v1` (not very obvious names and parameters, but it can be due to me not familiar with the API). What I meant by "a bit bigger and complicated" is that our Travis pipeline is currently 5 times smaller while having [almost] the same functionality.

> I could drop the 2nd commit and keep the travis build, so you can get a feel for it.

This is also an option indeed.

@agalakhov , @jxs what are your opinions?

I currently like our Travis file more just because it's much shorter and much more obvious. The GitHub Actions seem to be more sophisticated (perhaps by a price of bringing more control and flexibility?), but they also have some merits mentioned by @nickelc .

.github/workflows/ci.yml
agalakhov: Only build failures on 'stable' are hard errors.  Others are warnings.

agalakhov:
What's the profit of migrating?

application-developer-DA:
According to @nickelc's comment above:

> Better UI integration without leaving the website and workflows can run on forks without creating PRs.
It's also faster with the parallel execution of cargo test and the Autobahn testsuite.

.github/workflows/ci.yml
nickelc: GH lacks the `allow-failure` feature for specific build jobs but it supports branch protections in `Settings ` to require certain build jobs to pass before merging.

There is a workaround to use `continue-on-error` for  `beta/nightly` build steps to render them green but then you have to check the logs to see if these builds failed.

.github/workflows/ci.yml
nickelc: The matrix can only defined at job level and not for the whole workflow.
See [workflow syntax][workflow-syntax] for more information.

[workflow-syntax]: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategy

application-developer-DA:
Ok, let's try it!

PR #124, opened 2020-10-14
Update pin-project to 1

nickelc:


application-developer-DA (changes requested):
It would be nice to specify some more concrete version than just `1` ;)

PR #133, opened 2020-11-08
Fixed not to send Close frame at `poll_close`

emergent:
In version 0.11.0, the Close frame is automatically sent in the `poll_close` method of `Sink`. However, when using WebSocket, there are cases that the processing result from the server is received asynchronously with sending data from the client. (For example, in the case of voice recognition, the voice stream is sent and then the processing result of the server is received.)

For such a case, I thought that the process of automatically sending a Close frame should be removed from the crate and explicitly described on the application, so I modified it.

### DIFF
--- a/examples/client.rs
+++ b/examples/client.rs
@@ -58,4 +58,6 @@ async fn read_stdin(tx: futures_channel::mpsc::UnboundedSender<Message>) {
         buf.truncate(n);
         tx.unbounded_send(Message::binary(buf)).unwrap();
     }
+
+    tx.unbounded_send(Message::Close(None)).unwrap();
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -324,7 +324,7 @@ where
     }
 
     fn poll_close(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
-        match (*self).with_context(Some((ContextWaker::Write, cx)), |s| s.close(None)) {
+        match (*self).with_context(Some((ContextWaker::Write, cx)), |_| Ok(())) {
             Ok(()) => Poll::Ready(Ok(())),
             Err(::tungstenite::Error::ConnectionClosed) => Poll::Ready(Ok(())),
             Err(err) => {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -323,15 +323,8 @@ where
         (*self).with_context(Some((ContextWaker::Write, cx)), |s| cvt(s.write_pending()))
     }
 
-    fn poll_close(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
-        match (*self).with_context(Some((ContextWaker::Write, cx)), |_| Ok(())) {
-            Ok(()) => Poll::Ready(Ok(())),
-            Err(::tungstenite::Error::ConnectionClosed) => Poll::Ready(Ok(())),
-            Err(err) => {
-                debug!("websocket close error: {}", err);
-                Poll::Ready(Err(err))
-            }
-        }
+    fn poll_close(self: Pin<&mut Self>, _: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        Poll::Ready(Ok(()))
     }
 }
 
--- a/tests/communication.rs
+++ b/tests/communication.rs
@@ -94,7 +94,7 @@ async fn split_communication() {
     let tcp = TcpStream::connect("0.0.0.0:12346")
         .await
         .expect("Failed to connect");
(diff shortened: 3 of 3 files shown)


application-developer-DA:
Sorry, but I did not quite get why the change is required. If we close the connection, then we expect that the `Close` frame is sent. I can't imagine why someone would close the connection while expecting some additional data from the client at that point.

@strohel, what do you think?

application-developer-DA (changes requested):
Sorry, but I did not quite get why the change is required. If we close the connection, then we expect that the Close frame is sent. I can't imagine why someone would close the connection while expecting some additional data from the client at that point.

@strohel, what do you think?

agalakhov:
I'm pretty sure not sending `Close` at this time violates the RFC. Did you check it?

emergent:
@application-developer-DA @agalakhov,  thank you for your comments.

There are written in RFC6455 that

- The application MUST NOT send any more data frames after sending a Close frame.
- If an endpoint receives a Close frame and did not previously send a Close frame, the endpoint MUST send a Close frame in response.

But, there is not stated "the application must send a Close frame when you finish sending data".
There are cases where applications want to wait for the `Stream` result even after finishing the send channel (`Sink`).

The API of the speech recognition I mentioned in the first comment is the API we provide that receive the incoming voice stream, and respond the intermediate result asynchronously. When the audio stream has been sent, the client does not send a Close frame at that point. After receiving the final result of speech recognition, it receives a Close frame from the server and ends the session.

If an application send a Close frame immediately after sending the audio stream, the server wil return a Close frame without returning the speech recognition result and end the session. The RFC stipulates that this is the case.

I would like to write a client program for this API using `tokio-tungstenite` as follows:

https://github.com/emergent/rust-mimiio/blob/0d75d23d66e725d055201a0004b31ae2bcd80899/src/main.rs#L73-L91

However, with the current `Sink` implementation of `tokio-tungstenite`, a Close frame will be sent when the audio stream transmission is complete and the `Stream` will also end.

Therefore, the transmission channel must not be terminated until the reception on `Stream` is completed, so it must be implemented as follows.

https://github.com/emergent/rust-mimiio/blob/99d7d14a15c5e18ed89fb5da31e069e4a6ed5c7f/src/main.rs#L97-L101

I think that it is necessary for the application side to control the timing of sending the Close frame, rather than implementing it to suppress the process of automatically sending the Close frame.

What do you think?

application-developer-DA:
It took some time, but I think I finally can understand what is the problem you're experiencing.

First of all, the reason we can't simply "ignore" the fact that a user closes a sink (by not sending `Close` when `close()` is called) is because it would be logically incorrect: it's like having a file that you closed and then you expect that the `write()` method successfully writes something to this file. You can treat `poll_close()` as some sort of a destructor for the type that implements `Sink`: once `Sink` is teared down, it's considered closed and we can't write anything to it. Violating this rule would bring many issues.

Basically what are you trying to achieve is to be able to write to the sink which is dropped, but what I could not understand is why are you trying to do so, thing is: the `poll_close()` (which in turn sends the `Close` frame) is called by someone. So either you call `close()` manually on the sink (which would mean the problem is in an application logic, not in the `tokio-tungstenite`, because if you command the sink to close, then it will get closed) or the `poll_close()` or `close()` is called by something else which you're using in your code. In your particular case what I think must have happened is that you created some sort of a stream of events on the client and forwarded the stream to the WebSocket's `Sink` provided by `tokio-tungstenite`. The `Forward` is a `Stream` and this stream calls `poll_close()` on the `Sink` that it owns upon termination, closing the `Sink`, which is quite obvious and logical behavior considering that `forward()` takes `Sink` by value (i.e. moves it in).

In other words: you redirect the events from your event stream to the sink and at some point the original stream is closed, so is the `Sink`, which means that the `Close` frame is sent to the server (by the client), which in your particular case is not desired (because the server is meant to close the connection after processing all of the events?). If this is the case, then it's a problem with the application logic: if you want the WebSocket to be "writeable" after your stream of events terminates, then you should write it explicitly, i.e. you should not just forward the stream of events to the sink, assuming that the socket will be writable once sink terminates, because at the point sink terminates there is no way to send a close frame (as the `Sink` is dropped at this point).

What you could do is to re-think the logic of your application: i.e. maybe do something like this:
```
audio_events   ---
                  |===> Sink
control_events ---
```

I.e. the sink is not terminated by `audio_events`, but once both stream of events are finished, thus you still have a possibility to push of your `audio_events` and be able to send a `Close` from the client via `control_events` when you think it's appropriate.

I hope it makes sense.

application-developer-DA:
I hope that my comment helps you resolve the issue. I'm going to close this PR since it seems like changing the behavior will bring many undesired effects (including the inability to send `Close` from the client once the `WebSocket` is split into 2 parts where one part is consumed by some stream of events that forwards them to the sink).

PR #135, opened 2020-11-17
Regular maintenance + wanted to check how the new GitHub actions would work :P

application-developer-DA:


strohel (approved):
Nice! Automatisation FTW.

.github/workflows/ci.yml
```diff
@@ -35,6 +35,9 @@ jobs:
       - name: Test
         run: cargo test --release
 
+      - name: Format
+        run: cargo +nightly fmt
```
strohel, line 39: Actually formatting the code in actions doesn't probably make sense. We instead rather want to check whether formatting is okay:

```suggestion
        run: cargo +nightly fmt -- --check
```

application-developer-DA: Totally agree. I just forgot to use `--check` and wanted to see if `+nightly` would work at this place (discussed it with @jxs a week ago).

jxs (approved):

PR #146, opened 2021-01-01
Upgrade to tokio 1.0

coolreader18:
The interval server example needed updating, since `tokio::time::Interval` is no longer a stream; both the `.next()` and the `.tick()` futures are stateless, so it's okay to "create" them in each loop iteration.


coolreader18:
nevermind, I should have checked existing PRs first :sweat_smile:

PR #147, opened 2021-01-09
Add support for rustls as TLS backend, update tungstenite version

### LINKED ISSUE #127: Add tokio-rustls feature
All in the title, add a feature allowing use of rustls instead of native-tls, I'd be willing to work on this

dnaka91:
This adds support for `rustls` as TLS backend. It depends on `tungstenite-rs` 0.13.

Naming is exactly the same as in tungstenite and I added a missing feature for vendoring as well as it only existed in tungstenite so far: `native-tls-vendored`.

Fixes #127.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ stream = []
 log = "0.4"
 futures-util = { version = "0.3", default-features = false, features = ["async-await", "sink", "std"] }
 pin-project = "1.0"
-tokio = { version = "1.2.0", default-features = false, features = ["io-util"] }
+tokio = { version = "1.0.0", default-features = false, features = ["io-util"] }
 
 [dependencies.tungstenite]
 version = "0.13.0"
@@ -56,9 +56,9 @@ version = "0.21.0"
 
 [dev-dependencies]
 futures-channel = "0.3"
-tokio = { version = "1.2.0", default-features = false, features = ["io-std", "macros", "net", "rt-multi-thread", "time"] }
-url = "2.2.0"
-env_logger = "0.8.2"
+tokio = { version = "1.0.0", default-features = false, features = ["io-std", "macros", "net", "rt-multi-thread", "time"] }
+url = "2.0.0"
+env_logger = "0.7"
 
 [[example]]
 name = "autobahn-client"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ pin-project = "1.0"
 tokio = { version = "1.0.0", default-features = false, features = ["io-util"] }
 
 [dependencies.tungstenite]
-version = "0.12.0"
+version = "0.13.0"
 default-features = false
 git = "https://github.com/dnaka91/tungstenite-rs"
 branch = "rustls"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ stream = []
 log = "0.4"
 futures-util = { version = "0.3", default-features = false, features = ["async-await", "sink", "std"] }
 pin-project = "1.0"
-tokio = { version = "1.0.0", default-features = false, features = ["io-util"] }
+tokio = { version = "1.2.0", default-features = false, features = ["io-util"] }
 
 [dependencies.tungstenite]
 version = "0.13.0"
@@ -32,7 +32,7 @@ default-features = false
 [dependencies.native-tls-crate]
 optional = true
 package = "native-tls"
-version = "0.2.0"
+version = "0.2.7"
 
 [dependencies.rustls]
 optional = true
@@ -57,8 +57,8 @@ version = "0.21.0"
 [dev-dependencies]
 futures-channel = "0.3"
 tokio = { version = "1.2.0", default-features = false, features = ["io-std", "macros", "net", "rt-multi-thread", "time"] }
-url = "2.0.0"
-env_logger = "0.7"
+url = "2.2.0"
+env_logger = "0.8.2"
 
 [[example]]
 name = "autobahn-client"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,9 @@ edition = "2018"
 [features]
 default = ["connect"]
 connect = ["stream", "tokio/net"]
-tls = ["native-tls", "tokio-native-tls", "stream", "tungstenite/tls"]
+use-native-tls = ["native-tls", "tokio-native-tls", "stream", "tungstenite/use-native-tls"]
+use-native-tls-vendored = ["use-native-tls", "native-tls/vendored", "tungstenite/use-native-tls-vendored"]
+use-rustls = ["rustls", "tokio-rustls", "stream", "tungstenite/use-rustls", "webpki", "webpki-roots"]
 stream = []
 
 [dependencies]
@@ -26,14 +28,32 @@ tokio = { version = "1.0.0", default-features = false, features = ["io-util"] }
 [dependencies.tungstenite]
 version = "0.12.0"
 default-features = false
+git = "https://github.com/dnaka91/tungstenite-rs"
+branch = "rustls"
 
 [dependencies.native-tls]
 optional = true
 version = "0.2.0"
 
+[dependencies.rustls]
+optional = true
+version = "0.19.0"
+
 [dependencies.tokio-native-tls]
 optional = true
 version = "0.3.0"
+
+[dependencies.tokio-rustls]
+optional = true
+version = "0.22.0"
+
+[dependencies.webpki]
+optional = true
+version = "0.21.4"
+
+[dependencies.webpki-roots]
+optional = true
+version = "0.21.0"
 
 [dev-dependencies]
 futures-channel = "0.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,8 +28,6 @@ tokio = { version = "1.0.0", default-features = false, features = ["io-util"] }
 [dependencies.tungstenite]
 version = "0.13.0"
 default-features = false
-git = "https://github.com/snapview/tungstenite-rs.git"
-branch = "master"
 
 [dependencies.native-tls-crate]
 optional = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,9 +14,9 @@ edition = "2018"
 [features]
 default = ["connect"]
 connect = ["stream", "tokio/net"]
-use-native-tls = ["native-tls", "tokio-native-tls", "stream", "tungstenite/use-native-tls"]
-use-native-tls-vendored = ["use-native-tls", "native-tls/vendored", "tungstenite/use-native-tls-vendored"]
-use-rustls = ["rustls", "tokio-rustls", "stream", "tungstenite/use-rustls", "webpki", "webpki-roots"]
+native-tls = ["native-tls-crate", "tokio-native-tls", "stream", "tungstenite/native-tls"]
+native-tls-vendored = ["native-tls", "native-tls-crate/vendored", "tungstenite/native-tls-vendored"]
+rustls-tls = ["rustls", "tokio-rustls", "stream", "tungstenite/rustls-tls", "webpki", "webpki-roots"]
 stream = []
 
 [dependencies]
@@ -28,11 +28,12 @@ tokio = { version = "1.0.0", default-features = false, features = ["io-util"] }
 [dependencies.tungstenite]
 version = "0.13.0"
 default-features = false
-git = "https://github.com/dnaka91/tungstenite-rs"
+git = "https://github.com/dnaka91/tungstenite-rs.git"
 branch = "rustls"
 
-[dependencies.native-tls]
+[dependencies.native-tls-crate]
 optional = true
+package = "native-tls"
 version = "0.2.0"
 
 [dependencies.rustls]
@@ -57,6 +58,14 @@ version = "0.21.0"
 
 [dev-dependencies]
 futures-channel = "0.3"
-tokio = { version = "1.0.0", default-features = false, features = ["io-std", "macros", "rt-multi-thread", "time"] }
+tokio = { version = "1.2.0", default-features = false, features = ["io-std", "macros", "net", "rt-multi-thread", "time"] }
 url = "2.0.0"
 env_logger = "0.7"
+
+[[example]]
+name = "autobahn-client"
+required-features = ["connect"]
+
+[[example]]
+name = "client"
+required-features = ["connect"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,8 +28,8 @@ tokio = { version = "1.0.0", default-features = false, features = ["io-util"] }
 [dependencies.tungstenite]
 version = "0.13.0"
 default-features = false
-git = "https://github.com/dnaka91/tungstenite-rs.git"
-branch = "rustls"
+git = "https://github.com/snapview/tungstenite-rs.git"
+branch = "master"
 
 [dependencies.native-tls-crate]
 optional = true
--- a/src/connect.rs
+++ b/src/connect.rs
@@ -13,14 +13,14 @@ use tungstenite::{
 
 use super::{client_async_with_config, IntoClientRequest, WebSocketStream};
 
-#[cfg(feature = "use-native-tls")]
+#[cfg(feature = "native-tls")]
 pub(crate) mod encryption {
-    use native_tls::TlsConnector as NativeTlsConnector;
+    use native_tls_crate::TlsConnector as NativeTlsConnector;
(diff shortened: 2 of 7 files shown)


application-developer-DA (commented):
It also was released yesterday :)

dnaka91:
@application-developer-DA thanks for notifying me.

I updated the dependency and did a few more cleanups here and there. Should be ready for a review now.

src/connect.rs
```diff
@@ -41,29 +44,82 @@ pub(crate) mod encryption {
...
+
+    use crate::stream::Stream as StreamSwitcher;
+
+    /// A stream that might be protected with TLS.
+    pub type MaybeTlsStream<S> = StreamSwitcher<S, TlsStream<S>>;
+
+    pub type AutoStream<S> = MaybeTlsStream<S>;
+
+    /// A TLS connector that can be used when establishing TLS connections.
+    pub type TlsConnector = Arc<ClientConfig>;
```
strohel, line 78: There is a similar concern as in `tungstenite`: `MaybeTlsStream` and `TlsConnector` end up in public API sufrace of tokio-tungstenite. One crate may pull `tokio-tungstenite/rustls-tls`, and if another crate pulls `tokio-tungstenite/native-tls`, then the public API changes (possibly breaking the first crate).

The (reasonable) requirement that feature flags should be additive is really making our lives harder here.

We should probably try to somehow encapsulate this. Brainstorm: speaking about `MaybeTlsStream`, it could be also possible to implement the "dispatch" on Stream(Switcher), possibly making it `#[non_exhaustive]`. The question is whether such an encapsulation of the underlying stream would be limiting to our users - but I see that fields of Stream(Switch) are already private.

strohel (changes requested):
Thanks for your continued efforts, Dominik!

This looks nicely straightforward, though we may need an API surface treatment as in tungstenite. Please also add changelog and documentation for the new flags.

src/connect.rs
dnaka91: On seconds thought, maybe not that ready for review/merge yet 😅 

So after looking through the `MaybeTlsStream` and `TlsConnector` usages again I figured that both of them are quite troublesome.

- `MaybeTlsStream` can be extracted out of the `WebSocketStream` with the `get_ref` and `get_mut` functions. That means there is a chance that users may rely on the inner type being from `rustls` or `native-tls`.
- `TlsConnector` can be passed to `client_async_tls_with_config` which allows to customize the connector beforehand.

Encapsulating any of these two seems to either create a weird API or reduce customization options for the user. 

For example encapsulating `MaybeTlsStream` would either mean we hide the inner stream giving the user no access to its inner stream anymore. Or we use an enum but the users would need to match on it every time which is weird if you only have one TLS flag enabled (which is the case most of the time).

If we turn `TlsConnector` into an enum we would make it possible for users to pass the wrong connector to the function which would break things anyways. Say for example we pass a `rustls` connector but if both features are enabled it would use `native-tls` and the connector would not match so we would have to return an error anyways.

There is some way for sure to solve this nicely but currently I didn't find that way yet. I totally agree that the current implementation would break things, just not sure how to solve this well.

I will think about it a few days and see whether I can come up with something better.

Cargo.toml
```diff
@@ -14,29 +14,56 @@ edition = "2018"
...
+native-tls-vendored = ["native-tls", "native-tls-crate/vendored", "tungstenite/native-tls-vendored"]
+rustls-tls = ["rustls", "tokio-rustls", "stream", "tungstenite/rustls-tls", "webpki", "webpki-roots"]
 stream = []
 
 [dependencies]
 log = "0.4"
 futures-util = { version = "0.3", default-features = false, features = ["async-await", "sink", "std"] }
 pin-project = "1.0"
-tokio = { version = "1.0.0", default-features = false, features = ["io-util"] }
+tokio = { version = "1.2.0", default-features = false, features = ["io-util"] }
```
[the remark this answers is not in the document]
dnaka91: Oh sorry, that's my habit of updating versions. This works with 1.0.0 so I'll switch back the version number.
But as libraries don't have a lock file, how do we ensure that no new features from say 1.2.0 are currently used?

If the CI builds (or any other user) then cargo will create a Cargo.lock file with the latest version instead of 1.0.0. Somebody would need to change the version to `=1.0.0` from time to time (locally only) to see whether everything still works.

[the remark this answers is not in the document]
strohel: > But as libraries don't have a lock file, how do we ensure that no new features from say 1.2.0 are currently used?

This is a very valid concern. I don't really know what the best practices are here, or if some even exist regarding this matter. It is also possible that nobody actually uses compatible-but-older versions of libraries. :)

My comment was mainly in the terms of "a PR should only contain changes the connected with achieving the stated goal" to avoid scope creep.

src/connect.rs
strohel: Good points that the enum-based solution would be rather subpar.

Food for thought: we can side-step the problem by (conditionally) exposing 2 different streams: MaybeNativeTlsStream and MaybeRustTlsStream (or some shorter variant, or expose the generic parameter). We can help our users by making their API "textually the same", but not actually interchangeable in rustc sense. An analogy to std's BTreeMap vs. HashMap.

Optionally, we could provide some wrapper type on top of that for users that don't care about the particular implementation (majority?). Such a wrapper would come with the limitation that you cannot access the inner stream.

That's best I can think of right now. Feel encouraged to take your time and come up with own ideas - there's no need to hurry here.

Cargo.toml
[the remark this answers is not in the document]
dnaka91: That is very true. I will revert the other version upgrades as well then 🙇‍♂️

src/connect.rs
dnaka91: I made some rather big changes to the PR today, trying to solve this problem. Currently I'm quite happy with the result and found a solution where you can even control the used implemention with the `TlsConnector`.

I will do some more changes on it but if you want you can already have a look.

I'm thinking about whether the new `TlsStream` enum could be merged with the `MaybeTlsStream` 🤔

dnaka91: Maybe good to move the encryption module into its own file, but I would have to rename `connect.rs` to `mod.rs` and that makes the changes really hard to diff and review.

So for the sake of the review I would say better to do that as a separate PR afterwards.

Cargo.toml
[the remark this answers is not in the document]
strohel: > I will revert the other version upgrades as well then

It seems that these changes were not yet pushed?

src/connect.rs
```diff
@@ -5,95 +5,225 @@ use tokio::{
...
+    #[allow(missing_docs)]
+    #[non_exhaustive]
+    #[pin_project(project = StreamProj)]
+    pub enum TlsStream<S> {
+        Plain(PhantomData<S>),
+        #[cfg(feature = "native-tls")]
+        NativeTls(tokio_native_tls::TlsStream<S>),
+        #[cfg(feature = "rustls-tls")]
+        Rustls(tokio_rustls::client::TlsStream<S>),
+    }
```
strohel, line 40: Would it be possible to merge (functionality of) `TlsStream` and `crate::stream::Stream`? From a quick look, such a merge could prevent some logical duplication, and could avoid TlsStream::Plain being a dummy implementation doing nothing.

A natural name for such a merged type would be `MaybeTlsStream`. (i.e. it would take over the alias).

strohel (changes requested):
I've done only a partial review, but I like the direction this is taking. I'm submitting this review early so that we can set some high-level things first.

Cargo.toml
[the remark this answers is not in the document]
dnaka91: Yeah. With the latest change I was just focused on trying to find a nice solution for the feature conflicts. I'll add this together with the changelog additions soon.

src/connect.rs
dnaka91: Exactly my thought. 

I simply created new types as an experiment to see how well my idea works. In the end I saw that too but didn't have a chance to merged that with `crate::stream::Stream` yet.

Really like the idea of updating the name to `MaybeTlsStream`.

dnaka91:
Sorry @strohel I was a bit occupied this week and didn't find time to continue working on this.
I saw your mentions from the other issues/PRs.

Probably I will have some time tomorrow or on the weekend to complete the missing parts.

strohel:
> Sorry @strohel I was a bit occupied this week and didn't find time to continue working on this.
> I saw your mentions from the other issues/PRs.
> 
> Probably I will have some time tomorrow or on the weekend to complete the missing parts.

No problem at all, nothing is pressing to get this done quickly. (I also may be able to get more time to this only later next week).

dnaka91:
@strohel I merged the streams now and like how it looks now.

While doing so I noticed that the TLS features are behind the `connect` feature even though they actually don't have any relationship to the connect functions.
Therefore, I moved the TLS details into their own module so that you can use it even without `connect` enabled.

I know that this additional change is quite the stretch for this PR but 90% of the moved lines were new or modified anyways and I went with it.

If you don't like that I can roll back the last commit before adding the docs details, changelog update and Cargo.toml version reverts.

src/lib.rs
```diff
@@ -17,7 +17,9 @@ mod compat;
 mod connect;
 mod handshake;
 #[cfg(feature = "stream")]
-pub mod stream;
+pub(crate) mod stream;
```
strohel, line 20: nit: err, does the `pub(crate)` make any difference for these top-level items?

src/stream.rs
```diff
@@ -11,44 +11,57 @@ use std::{
...
+    /// Encrypted socket stream using [`native-tls`].
+    #[cfg(feature = "native-tls")]
+    NativeTls(tokio_native_tls::TlsStream<S>),
+    /// Encrypted socket stream using [`rustls`].
+    #[cfg(feature = "rustls-tls")]
+    Rustls(tokio_rustls::client::TlsStream<S>),
 }
 
-impl<S: AsyncRead + Unpin, T: AsyncRead + Unpin> AsyncRead for Stream<S, T> {
+impl<S: AsyncRead + AsyncWrite + Unpin> AsyncRead for MaybeTlsStream<S> {
```
strohel, line 27: Just curious: is AsyncWrite really needed for implementing AsyncRead? (converse case below)

src/stream.rs
```diff
@@ -11,44 +11,57 @@ use std::{
 
 use tokio::io::{AsyncRead, AsyncWrite, ReadBuf};
 
-/// Stream, either plain TCP or TLS.
+/// A stream that might be protected with TLS.
 #[pin_project(project = StreamProj)]
-pub enum Stream<S, T> {
+pub enum MaybeTlsStream<S> {
```
strohel, line 16: I think we should add `#[non_exhaustive]`.

src/tls.rs
```diff
@@ -0,0 +1,194 @@
...
+    client::uri_mode, error::Error, handshake::client::Response, protocol::WebSocketConfig,
+};
+
+use crate::{client_async_with_config, domain, IntoClientRequest, WebSocketStream};
+
+pub use crate::stream::MaybeTlsStream;
+
+/// A TLS connector that can be used when establishing TLS connections.
+#[non_exhaustive]
+pub enum TlsConnector {
```
strohel, line 14: brainstorm: maybe `Connector` name would be more appropriate when one variant is `Plain`?

src/tls.rs
```diff
@@ -0,0 +1,194 @@
+//! Connection helper.
+use tokio::io::{AsyncRead, AsyncWrite};
+
+use tungstenite::{
+    client::uri_mode, error::Error, handshake::client::Response, protocol::WebSocketConfig,
+};
+
+use crate::{client_async_with_config, domain, IntoClientRequest, WebSocketStream};
+
+pub use crate::stream::MaybeTlsStream;
```
strohel, line 10: Hmm, MaybeTlsStream seems only available under `stream` flag, but here it is imported unconditionally - I wonder whether this compiles with `*-tls` + `!stream`...?

src/tls.rs
```diff
@@ -0,0 +1,194 @@
...
+    NativeTls(native_tls_crate::TlsConnector),
+    /// `rustls` TLS connector.
+    #[cfg(feature = "rustls-tls")]
+    Rustls(std::sync::Arc<rustls::ClientConfig>),
+}
+
+mod encryption {
+    #[cfg(feature = "native-tls")]
+    pub mod native_tls {
+        use native_tls_crate::TlsConnector;
```
strohel, line 28: I suggest either the `TlsConnector` -> `Connector` renama above, or importing this as `NativeTlsConnector` - it's confusing that this name refers to different types throughout the file.

strohel (changes requested):
I like the direction this has taken! I've attached some questions.

strohel:
Please also rebase on current master. It would be cool if you also addressed the older comments - it's most easy for me to review things that are "merge-ready". Thanks!

src/stream.rs
dnaka91: Unfortunately yes, due to `rustls`. Here is the signature of rustls's `TlsSTream`:
```rust
impl<IO> AsyncRead for TlsStream<IO>
where
    IO: AsyncRead + AsyncWrite + Unpin
```

src/stream.rs
dnaka91: Definitely, I forgot this one. Thanks for noting 🙇

src/tls.rs
dnaka91: That's a very good idea. Not only does it repeat the module name (`tls::TlsConnector`), but it probably would raise questions what a plain TLS connector is ought to be.

I'll rename to simply `Connector`.

src/tls.rs
dnaka91: This is save as you can't disable the `stream` feature and have a `*-tls` feature at the same time.
The `*-tls` features require the `stream` feature so the developer can't disable it.

In addition I didn't get any errors when running `cargo hack check --feature-powerset --no-dev-deps` which tests all possible feature flag combinations.

src/tls.rs
dnaka91: That is definitely true. I decided to rename the crate's connector to simply `Connector` so this should be fine then.
I can additionally rename the import to `NativeTlsConnector` to reduce confusion if you like?

src/lib.rs
dnaka91: No differences there. I did this together with the connect/tls split to have a nicer public API.
The only thing that's available from the stream module is the `MaybeTlsStream` and that's now re-exported on the top-level for a simpler API structure.

I can of course revert this back if you prefer to have this module public and the `MaybeTlsStream` not at the top-level.

strohel: Oh, I expressed myself to vaguely. I like `MaybeTlsStream` being promoted to the top-level.

I meant this simple change (without any intended change of functionality):
```suggestion
mod stream;
```
That seems to work? Same for `mod tls`.

src/stream.rs
strohel: I see. Not a real problem.

src/tls.rs
strohel: > This is save as you can't disable the stream feature and have a *-tls feature at the same time.
The *-tls features require the stream feature so the developer can't disable it.

Oh, that's the missing piece that didn't occur to me, thanks for explaining. It seemed strange to me because I know you're running `cargo hack`.

src/tls.rs
strohel: > I decided to rename the crate's connector to simply Connector so this should be fine then.

:+1: 

> I can additionally rename the import to NativeTlsConnector to reduce confusion if you like?

Up to you, `NativeTlsConnector` is nicely explicit. But no longer necessary to rename it due to the other rename.

strohel (approved):
This looks good! Just one super-nit left.

I think now it is a good time that somebody less in-the-box looks at this. @application-developer-DA, @jxs?

application-developer-DA (approved):

jxs (approved):

application-developer-DA:
Ok, merging. Thanks @dnaka91!

strohel:
Thank you for the second time, @dnaka91. The whole Rust team here at Snapview admires the diligence that you've put into these PRs. :heart:

(we may have merged a tiny bit too early without giving you a chance to do last little fixups, feel free to open new PRs if you like)

dnaka91:
Thank you for the kind words @strohel ❤️ 

Yeah that was merged a little quick in the end so I'll change that `pub(crate) mod stream` in a separate tiny PR, no problem.

strohel:
> Yeah that was merged a little quick in the end so I'll change that pub(crate) mod stream in a separate tiny PR, no problem.

@application-developer-DA was so quick he did it in https://github.com/snapview/tokio-tungstenite/commit/c3e76eae904af27d5aceb2950d58c05cfb67632d already. :)

PR #149, opened 2021-01-18
depend on tokio 1.0 instead of 1.0.0; fixes #148

### LINKED ISSUE #148: Using tokio 1.0.0 which has been yanked
https://crates.io/crates/tokio/versions

We could relax this dependency to tokio 1.0

I'll put in a PR

paulirotta:
Change `Cargo.toml` to depend on tokio 1.0

fixes #148

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,6 @@ version = "0.3.0"
 
 [dev-dependencies]
 futures-channel = "0.3"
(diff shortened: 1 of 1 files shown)


strohel:
You were quicker than my comment in https://github.com/snapview/tokio-tungstenite/issues/148: "1.0" and "1.0.0" are equivalent.

paulirotta:
No actual effect

PR #153, opened 2021-02-15
Update tungstenite version

hei-pa:
Please update the tungstenite version to support current stable version


strohel:
Hi @hei-pa, thanks for your effort. This is already being worked on in https://github.com/snapview/tokio-tungstenite/pull/147, so I'll proceed to close this PR.

I've changed the title of the other PR so that it is more discoverable in the pull request list (CC @dnaka91).

PR #156, opened 2021-02-23
Use 127.0.0.1 in tests to avoid firewall warnings

dnaka91:
As the title says, this changes all tests to run on 127.0.0.1 instead of 0.0.0.0.

When for example on MacOS you get a firewall warning for every time the code changes and the tests are run as the binary changed. Some firewalls even let the tests hang so they have to be cancelled and restarted several times.

By changing to 127.0.0.1 this warnings don't come as the binary won't listen for requests from external devices.

### DIFF
--- a/src/handshake.rs
+++ b/src/handshake.rs
@@ -92,11 +92,11 @@ pub(crate) async fn client_handshake<F, S>(
 ) -> Result<(WebSocketStream<S>, Response), Error<ClientHandshake<AllowStd<S>>>>
 where
     F: FnOnce(
-        AllowStd<S>,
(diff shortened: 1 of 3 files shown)


strohel (approved):
LGTM, thanks.

It looks like the rustfmt CI has just switched to a new version which wants to do some formatting. Feel free to fix that in a separate commit if you want.

dnaka91:
Ah this were the same that I discovered in my `rustls` branch. FIxed it 🙇‍♂️

application-developer-DA (approved):

PR #158, opened 2021-03-02
Enable `native-tls` and `rustls-tls` features on docs.rs

nickelc:
Current documentation is missing the types and functions for tls.


dnaka91:
This is great, thanks 👍 
I just went into the docs a moment ago and realized that I forgot to add the feature flags for docs.rs in #147.

PR #164, opened 2021-04-17
Add documentation about TLS features to README

torlenor:
Because I was struggling with this and did not think about looking into the Cargo.toml for the features, I would like to add a few lines to the README describing the features required for TLS support.


strohel (approved):
Thanks, this is perfect.

The autobahn test failures look unrelated. Do you have a clue, @application-developer-DA @jxs?

application-developer-DA:
Yes, it looks like Autobahn Test Suite does not work anymore when we try to start it. It could be that the default `pip` is now `pip3` which downloads the suite that does not run with Python 3. I'll try to look into that.

PR #165, opened 2021-04-21
Fix Autobahn Scripts

application-developer-DA:
See commit messages for more details.

**NOTE:** I also noticed that one of the files with results for Autobahn was [apparently blindly] copied in our repositories and is not used neither in `tungstenite-rs` nor in `tokio-tungstenite`. I tried to include this small clean up in this PR (could not hold myself leaving it as is), but I plan to submit a separate PR to do the same in `tungstenite-rs`. Additionally, there is a possibility to remove the code duplication for the tests that we do for Autobahn and to extract the common part, so that `tokio-tungstenite` and `tungstenite-rs` do not need to contain the same code.


scripts/autobahn-client.sh
```diff
@@ -23,10 +18,12 @@ function test_diff() {
     fi
 }
 
-cargo build --release --example autobahn-client
+docker run -d --rm \
+    -v "${PWD}/autobahn:/autobahn" \
+    -p 9001:9001 \
+    crossbario/autobahn-testsuite \
+    wstest -m fuzzingserver -s 'autobahn/fuzzingserver.json'
```
strohel, line 24: We start the container in the background (using `-d`), but AFAICS we never tear down the container.

Instead of removing the `function cleanup()` and `trap`, we should update them to tear down the container. `docker run -d` returns container ID, which can be assigned to a variable for that use-case.

scripts/autobahn-server.sh
```diff
@@ -23,9 +22,13 @@ function test_diff() {
...
 
-cargo build --release --example autobahn-server
 cargo run --release --example autobahn-server & WSSERVER_PID=$!
-echo "Server PID: ${WSSERVER_PID}"
 sleep 3
-wstest -m fuzzingclient -s 'autobahn/fuzzingclient.json'
+
+docker run -d --rm \
+    -v "${PWD}/autobahn:/autobahn" \
+    --network host \
```
strohel, line 30: As an alternative to `--network host`, we can/should use `172.17.0.1` to contact the Docker host machine from within the container. (though this may not be a portable solution - see https://stackoverflow.com/questions/22944631/how-to-get-the-ip-address-of-the-docker-host-from-inside-a-docker-container)

strohel (changes requested):
Looks good, just some technicalities.

BTW, it seems that GH Actions work well - we can as well remove Travis, if it has been completely superseded by GH Actions.

scripts/autobahn-client.sh
application-developer-DA: Ok, that's a valid concern that I had, but I decided to not do it, because I assumed that once the job ends, so do all the processes, including docker and its containers. Because even if I stop the container (just for the sake of not having some containers running at the end of the job), it is still be there (it won't be pruned automatically).

> Applies also to scripts/autobahn-server.sh.

Does it? I'm asking because in case of `autobahn-server` we start our own server (that is covered by the `cleanup()`) and we use Autobahn only for the fuzzing client. Once the client process ends, the container is stopped. The same for

scripts/autobahn-server.sh
application-developer-DA: Yes, the portability was one of my concerns. That would also mean that the config must be adjusted, so that the fuzzingclient is configuredf to connect to `172.17.0.1`. Not a big deal provided we use it only in conjunction with docker in this particular configuration.

scripts/autobahn-client.sh
strohel: > Ok, that's a valid concern that I had, but I decided to not do it, because I assumed that once the job ends, so do all the processes, including docker and its containers.

I was thinking not just about the CI, but more about people that would run that locally (to test tungstenite changes/to test new autobah/you testing this PR). And just as a general good practice to clean up resources that are created.

> Because even if I stop the container (just for the sake of not having some containers running at the end of the job), it is still be there (it won't be pruned automatically).

This looks like a problem, because we specify `--rm`, so the container should be destroyed. Let me guess: the container suffers from the "PID 1" problem, i.e. the CTRL+C signal is not caught by it properly, so that it is not stopped right away after `docker stop` and is killed only after the graceful stop timeout (10s by default). If this is the case, mitigation is easy: add `--init` to `docker run` flags.

(reply to the other part in other code comment)

scripts/autobahn-server.sh
```diff
@@ -14,18 +13,22 @@ trap cleanup TERM EXIT
...
     fi
 }
 
-cargo build --release --example autobahn-server
 cargo run --release --example autobahn-server & WSSERVER_PID=$!
-echo "Server PID: ${WSSERVER_PID}"
 sleep 3
-wstest -m fuzzingclient -s 'autobahn/fuzzingclient.json'
+
+docker run -d --rm \
```
strohel, line 28: > > Applies also to scripts/autobahn-server.sh.
> Does it? I'm asking because in case of autobahn-server (...) we use Autobahn only for the fuzzing client. Once the client process ends, the container is stopped.

Oh I see, you're right.

In this case, we should _not_ add `-d` to `docker run` flags here. The `-d` causes that `docker run` command is non-blocking, so `test_diff` below would be run too early in that case IMO. (I wonder how that could work in current PR state).

scripts/autobahn-server.sh
strohel: I see, okay, feel free to keep `--network host` and resolve this thread, it is not a big deal.

Note that `--network host` is reportedly even less portable, per https://docs.docker.com/network/host/:
> The host networking driver only works on Linux hosts, and is not supported on Docker Desktop for Mac, Docker Desktop for Windows, or Docker EE for Windows Server.

(though some my friends say that Docker host networking works for them as expected on Macs)

scripts/autobahn-client.sh
application-developer-DA: > If this is the case, mitigation is easy: add --init to docker run flags.

Oh nice, thanks for the info, I was afraid that the whole `dumb-init` thing would be required.

scripts/autobahn-server.sh
application-developer-DA: Oh, alright. Yeah, seems to work on my M1 Mac as well...

application-developer-DA:
> BTW, it seems that GH Actions work well - we can as well remove Travis, if it has been completely superseded by GH Actions.

Yeah, I also thought about that. I still enjoy the elegance and conciseness of Travis (compare it to GitHub Actions job definition that is much longer), but having both at the same time does not seem like a reasonable use of resources.

Though we don't use Github Actions in `tungstenite-rs`, only in `tokio-tungstenite`, so I'll probably just remove Travis from here and leave if in `tungstenite-rs` (allows to enjoy and compare both).

strohel (approved):
LGTM now, feel free to merge.

PR #179, opened 2021-06-24
Fixed the "client.rs" example (use "futures" crate, etc.)

maanbsat:


jxs:
Hi, and thanks! Can you look into the errors reported by the CI?
thanks

application-developer-DA:
Could you please elaborate on why the change in the code was required? The examples did compile well even before this change.

Unfortunately there are a couple issues with this PR:
1. There is an additional "Merge branch" commit which is not necessary, we need a clean commit history ;)
2. The example changes the crates being used without changing the `[dev-dependencies]`, so once we merge this PR the client example **won't compile** anymore when executing it with `cargo run --example`.
3. It's not clear why we should use a different set of crates/dependencies only for the client example while using a different one for the rest of them. I think we ideally need some consistency.

application-developer-DA:
We decided to close this PR since there has been no elaboration on it for a couple of months, so we're not sure there will be any updates on this.

PR #183, opened 2021-08-09
Add a function to connect with a specific connector

Erk-:
This makes reusing a tls connector much easier as you will not have to
set up the socket to connect on yourself. When using native-tls this
can save a considerable amount of memory.


application-developer-DA (approved):

jxs (approved):
LGTM!

PR #185, opened 2021-08-11
Reduce crate package size

dnaka91:
Same as in https://github.com/snapview/tungstenite-rs/pull/229, reducing crate package size to improve download speed for everyone.

Especially big reductions here due to the removal of autobahn JSON files from the package.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,7 @@ documentation = "https://docs.rs/tokio-tungstenite/0.15.0"
 repository = "https://github.com/snapview/tokio-tungstenite"
 version = "0.15.0"
 edition = "2018"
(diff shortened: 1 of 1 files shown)


jxs:
Hi, and thanks! One question though, why do we need the `examples` dir on this case, and not on https://github.com/snapview/tungstenite-rs/pull/229 ?

dnaka91:
That is because the `Cargo.toml` references some examples to define their feature flag requirements. Sadly it seems to need the examples in `cargo package` to verify the project.

They're not needed in `tungstenite-rs` because there are no configs for the examples present.

jxs:
ah I see, thanks for clarifying :grin:

jxs (approved):

Issue #188, opened 2021-09-06
Echo server panicked with ConnectionClosed error


Leo1003:
## Information
- Version: master (d566387)
- OS: Arch Linux
- Rust: 1.54.0

## Description

When I tried to run the echo server sample and using `websocat` to communicate with it.
The `accept_connection()` panicked with the `ConnectionClosed` error when the websocat reached EOF and closed the connection gracefully.

<details>
  <summary>Stack trace</summary>

```
thread 'tokio-runtime-worker' panicked at 'Failed to forward message: ConnectionClosed', examples/echo-server.rs:45:31
stack backtrace:
   0: rust_begin_unwind
             at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/std/src/panicking.rs:515:5
   1: core::panicking::panic_fmt
             at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/core/src/panicking.rs:92:14
   2: core::result::unwrap_failed
             at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/core/src/result.rs:1355:5
   3: core::result::Result<T,E>::expect
             at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/core/src/result.rs:997:23
   4: echo_server::accept_connection::{{closure}}
             at ./examples/echo-server.rs:45:5
   5: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/core/src/future/mod.rs:80:19
   6: <core::pin::Pin<P> as core::future::future::Future>::poll
             at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/core/src/future/future.rs:119:9
   7: tokio::runtime::task::core::CoreStage<T>::poll::{{closure}}
             at /home/leo/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.11.0/src/runtime/task/core.rs:161:17
   8: tokio::loom::std::unsafe_cell::UnsafeCell<T>::with_mut
             at /home/leo/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.11.0/src/loom/std/unsafe_cell.rs:14:9
   9: tokio::runtime::task::core::CoreStage<T>::poll
[...]

PR #189, opened 2021-09-10
Fix the echo server example to prevent ConnectionClosed error

Leo1003:
Related #188

The echo server would generate ConnectionClosed error when the peer close the connection normally. 

This is because that `tungstenite` crate automatically reply Close message to the peer. 
Then `StreamExt::forward()` also forwards the Close message after the connection is closed.

### DIFF
--- a/examples/echo-server.rs
+++ b/examples/echo-server.rs
@@ -10,7 +10,7 @@
 
 use std::{env, io::Error};
 
-use futures_util::StreamExt;
+use futures_util::{future, StreamExt, TryStreamExt};
 use log::info;
 use tokio::net::{TcpListener, TcpStream};
 
@@ -42,5 +42,9 @@ async fn accept_connection(stream: TcpStream) {
     info!("New WebSocket connection: {}", addr);
 
     let (write, read) = ws_stream.split();
-    read.forward(write).await.expect("Failed to forward message")
+    // We should not forward messages other than text or binary.
+    read.try_filter(|msg| future::ready(msg.is_text() || msg.is_binary()))
+        .forward(write)
+        .await
+        .expect("Failed to forward messages")
 }


examples/echo-server.rs
```diff
@@ -41,6 +42,16 @@ async fn accept_connection(stream: TcpStream) {
 
     info!("New WebSocket connection: {}", addr);
 
-    let (write, read) = ws_stream.split();
-    read.forward(write).await.expect("Failed to forward message")
+    let (mut write, mut read) = ws_stream.split();
+    while let Some(msg) = read.next().await {
+        let resp = match msg.expect("Failed to read messages") {
+            msg @ (Message::Text(_) | Message::Binary(_)) => msg,
+            Message::Ping(data) => Message::Pong(data),
```
jxs, line 49: we don't need to send a `Pong` as `tungstenite` [automatically does it for us](https://github.com/snapview/tungstenite-rs/blob/master/src/protocol/mod.rs#L172-L175)

jxs (changes requested):
thanks! Only some suggestion left

examples/echo-server.rs
Leo1003: Oh, I see. Thanks for pointing it out.

examples/echo-server.rs
```diff
@@ -41,6 +42,15 @@ async fn accept_connection(stream: TcpStream) {
...
+    while let Some(msg) = read.next().await {
+        let resp = match msg.expect("Failed to read messages") {
+            msg @ (Message::Text(_) | Message::Binary(_)) => msg,
+            // Match Message::Close here.
+            // We should not send any message after the connection is closed.
+            Message::Close(_) => break,
+            _ => continue,
+        };
+        write.send(resp).await.expect("Failed to send messages");
+    }
```
application-developer-DA, line 55: Sorry for the late review.

I think there is a bit more elegant way to write it preserving the previous `read.forward(write)` notation with combinators, e.g. something like `read.filter(|m| m.is_binary() || m.is_text()).forward(write)`.

NB: Though you'll have to use `try_filter` I believe because of the type of the `Stream`, but I think you got the point :)

application-developer-DA (changes requested):

examples/echo-server.rs
Leo1003: Fixed with an elegant way! It looks nicer now.

application-developer-DA (approved):

application-developer-DA:
Thanks!

PR #190, opened 2021-09-13
Add `permessage-deflate` support

kazk:
Adds `WebSocketStream::from_raw_socket_with_compression` to enable compression without handshake for crates like `warp` and `axum`.

See https://github.com/snapview/tungstenite-rs/pull/235

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 target
 Cargo.lock
+autobahn/client/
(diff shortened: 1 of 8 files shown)


kazk:
Superseded by #251. If anyone is using this branch, I'm not deleting it because I also have some projects depending on it.

Issue #200, opened 2021-11-07
Semantics of stream read yielding `Ok(None)` and `Err(e)`?


wbrickner:
I am confused on some of the semantics surrounding streams
yielding `Option<Result<...>>` in `tokio-tungsteninte`. 

Let's say I have a setup like this:

```rust
let (stream, _) = connect_async( /* ... */ ).await;
let (_, mut rx) = socket.split();

loop {
  match read.next().await {
    Some(Ok(msg)) => { /* ... */ },

    // The stream has not ended, but there was an error with reading from the socket,
    // but an error not sufficient to close the stream? What?
    Some(Err(e)) => { },

    // The stream has ended... so this is a distinct type of error with the socket? What?
    None => {  }
  }
}
```

I have witnessed streams yield `Some(Err(Protocol(...)))` and `None` (I think separately).

- What are the conditions in which `read.next().await` yields `None`? I want to know what this means must have happened.
- Why can a stream yield an error (e.g. a `Protocol` error – seemingly fatal) and not be ended (by a `None`)? Is this just an artifact of the stream abstraction / type design?

Thanks so much,
I really enjoy using your library in my daily work, I hope you can help me with my question.

application-developer-DA:
So the thing is that every stream must eventually end. Like every iterator, when the iterator ends, it returns `None`, you can view streams as an abstract/future-aware version of a normal iterator from the standard library. So the stream retuning `Option` is a generic requirement for all streams that we had to implement which also allows all that "interop" with other `StreamExt` functions, allowing `for_each`, `fold`, `forward` etc.

This also means that if the connection is ended gracefully on either of the sides, we don't really want to return any error, but "gracefully" close the stream (by returning `None`). In `tungstenite-rs` (core) the function that deal with reading return `Result<X, Error>`, where `Error::ConnectionClosed` means a normal close of the connection. What we do on `tokio-tungstenite` is we turn `Error::ConnectionClosed` into `None` on the stream, signifying a good graceful closure of the connection.

[...]

PR #221, opened 2022-03-22
Make the semantics of `Stream` more reasonable

application-developer-DA:
Closes https://github.com/snapview/tokio-tungstenite/issues/200.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -183,6 +183,7 @@ where
 pub struct WebSocketStream<S> {
     inner: WebSocket<AllowStd<S>>,
     closing: bool,
+    ended: bool,
 }
 
 impl<S> WebSocketStream<S> {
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -271,13 +272,27 @@ where
 
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
         trace!("{}:{} Stream.poll_next", file!(), line!());
+
+        // The connection has been closed or the critical error has occurred.
+        // We have already returned the error to the user, the `Stream` is unusable,
+        // so we assume that the stream has been "fused".
```
sdroege, line 278: You probably also want to implement `futures::stream::FusedStream` then

application-developer-DA: Oh, great suggestion! Thanks. Will update the PR right away.

sdroege (approved):

PR #226, opened 2022-04-25
These helper functions allow a socket to be supplied to the connect methods

john-sharratt:


application-developer-DA:
Sorry, but I don't see how this could be useful: we provide the `client_async_tls_with_config` that already accepts the stream (that could well be a `TcpStream` or something else), so I don't see how the addition of a new function would help. Unfortunately, I think it's rather confusing than useful (e.g. a `connect*` function that does not perform a connect, but merely forwards the stream to the `client_async*` function.

john-sharratt:
@application-developer-DA that function is in a private crate (tls) so it can't be called - at least I didn't see a way

john-sharratt:
Ah found it! My bad
Thanks for helping

PR #229, opened 2022-05-20
Don't consider it an error if the connection is closed while flushing

Liamolucko:
It's perfectly fine for the connection to close while flushing if the thing you're flushing is a responding close frame.
I'm pretty sure that there isn't any other scenario where `tungstenite` will return `Error::ConnectionClosed` when writing.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -334,10 +334,8 @@ where
     fn poll_flush(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
         (*self).with_context(Some((ContextWaker::Write, cx)), |s| cvt(s.write_pending())).map(
             |result| match result {
-                // It's not an error for the WebSocket to have closed normally.
-                // Note: `tungstenite` will only return this if a close frame has been sent, and
-                // doesn't allow sending anything else after that, so this can't arise when trying
-                // to flush something other than a close frame.
+                // It's not an error for the WebSocket to close while flushing if we're supposed to
+                // be closing - that just means that our close frame got flushed.
                 Err(::tungstenite::Error::ConnectionClosed) if self.closing => Ok(()),
                 other => other,
             },
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -338,7 +338,7 @@ where
                 // Note: `tungstenite` will only return this if a close frame has been sent, and
                 // doesn't allow sending anything else after that, so this can't arise when trying
                 // to flush something other than a close frame.
-                Err(::tungstenite::Error::ConnectionClosed) => Ok(()),
+                Err(::tungstenite::Error::ConnectionClosed) if self.closing => Ok(()),
                 other => other,
             },
         )
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -332,7 +332,16 @@ where
     }
 
     fn poll_flush(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -332,7 +332,16 @@ where
...
 
     fn poll_flush(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
-        (*self).with_context(Some((ContextWaker::Write, cx)), |s| cvt(s.write_pending()))
+        (*self).with_context(Some((ContextWaker::Write, cx)), |s| cvt(s.write_pending())).map(
+            |result| match result {
+                // It's not an error for the WebSocket to have closed normally.
+                // Note: `tungstenite` will only return this if a close frame has been sent, and
+                // doesn't allow sending anything else after that, so this can't arise when trying
+                // to flush something other than a close frame.
+                Err(::tungstenite::Error::ConnectionClosed) => Ok(()),
```
sdroege, line 341:
```suggestion
                Err(::tungstenite::Error::ConnectionClosed) if self.closing => Ok(()),
```

If the connection is closed at another time while flushing then that seems problematic. Flushing doesn't only happen when closing.

Liamolucko: It doesn't seem like `self.closing` gets set in all the cases where the `WebSocket` gets closed - it's set in `poll_close`, but not when a close message is sent or recieved. (The case where a close message is received is the case where this situation should actually occur, since we're flushing the responding close frame automatically sent by `tungstenite`.)

sdroege (approved):

src/lib.rs
sdroege: I guess that needs some updating too then

application-developer-DA: The semantics of `closing` was that it was set only in cases when the user calls `poll_close()` for the following reason: our initial implementation (written years ago) did only one thing when `poll_close()` was called: it essentially added the `Close` message to the queue and tried to write it but it did not ensure that the `Close` frame is flushed (is actually sent to the peer). [There was a PR filed back then](https://github.com/snapview/tokio-tungstenite/pull/215) that fixed this behavior by making sure that once `poll_close()` is called, we send the `Close` frame and the subsequent calls to the `poll_close()` ensure that we're just flushing it ensuring that the queue has been flushed.

In other words to trigger the behavior introduced in this commit, one must first call `poll_close()`, get `Pending`, and then call `poll_flush()` (I don't know if it's the proper scenario? I assumed that normally when one calls `poll_close()` and it's not ready, one is supposed to call it until it's closed). I'm not sure if calling the `poll_flush()` after `poll_close()` is a good idea (is it?).

In other words: if you never call `poll_close()`, you'll never reach the branch of the code that this PR changes.

daniel-abramov:
The author seems to be right, we seem to have returned `ConnectionClosed` from `poll_flush()` which should not be the case (since `poll_flush()` is meant to just flush the queue, and the `ConnectionClosed` returned from there just indicates that the connection has just been closed, i.e. the flush has been completed).

So it seems like we simply could ignore `ConnectionClosed` in `poll_flush()` if I'm not mistaken.

This has been addressed here: https://github.com/snapview/tokio-tungstenite/pull/281
And then I also removed the boilerplate (took over this part from this PR): https://github.com/snapview/tokio-tungstenite/commit/cb1196967676a71a885faa9188b695d95f456ee7

Hopefully, we'll now get a more predictable and simple behavior 🙂 (we would ideally need to revisit the state machine in `tungstenite` someday since it has gotten a bit more complicated than it probably should)

PR #237, opened 2022-07-31
Add an example of using tungstenite with a custom accept.

ahmedcharles:


application-developer-DA:
Hm, interesting way of using it bypassing `tungstenite`'s handshake mechanism and making it via `hyper`. Would you please address the issue with `cargo fmt` that currently fails the pipeline?

ahmedcharles:
It should be good now. I hope it implements the same checks that tungstenite normally does. Thanks.

PR #252, opened 2023-01-15
connect: disable Nagle algorithm

### LINKED ISSUE #250: Can I set_nodealy(true) for the tcpstream?
https://github.com/snapview/tokio-tungstenite/blob/05353b728ecff31a98feb6a75d404f648c5d5410/src/connect.rs#L49

I want to disable TCP Nagle Algorithm.

daniel-abramov:
Fixes #250.

### DIFF
--- a/src/connect.rs
+++ b/src/connect.rs
@@ -48,6 +48,9 @@ where
     let try_socket = TcpStream::connect(addr).await;
     let socket = try_socket.map_err(Error::Io)?;
 
+    // Disable Nagle algorithm.
+    socket.set_nodelay(true)?;
+
     crate::tls::client_async_tls_with_config(request, socket, config, None).await
 }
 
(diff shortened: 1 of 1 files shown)


sdroege:
That should probably be configurable? I don't think `TCP_NODELAY` is always the right choice. See e.g. recent https://withinboredom.info/blog/2022/12/29/golang-is-evil-on-shitty-networks/ , or just searching the Internet for it will show lots of people running into problems in various scenarios where that broke things.

daniel-abramov:
@sdroege, that's a good point (as always)! I've read the article you provided and tried to refresh my knowledge of the Nagle algorithm and indeed it seems like having the `TCP_NODELAY` off is not that bad. I remember that in `tungstenite` we had the `TCP_NODELAY` enabled primarily of the concerns that the smaller packets won't be sent immediately and would cause not only delays, but problems on the receiver side that may not get a message at all if it's very small, but after re-checking the Nagle's algorithm I've figured that it's not the case as it seems like it would buffer the smaller messages only if there are unacknowledged segments while sending them otherwise.

x1957:
LGTM

gautamg795:
hi, thanks for making this change! would love to get the next release with this change pushed to crates.io so we can start using it in some projects soon :)

PR #254, opened 2023-01-24
Support UNIX domain sockets

da-x:


daniel-abramov:
Agree with @agalakhov, I think it's one of those things that seem to be a better fit for a user code (after all, it's a small function and just one of the use-cases where the WebSockets could be used).

PR #262, opened 2023-02-08
Improve CI build jobs

nickelc:
The Ubuntu images already come with docker and `cargo hack` is now installed with [`taiki-e/install-action`](https://github.com/taiki-e/install-action).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,7 @@ jobs:
         run: sudo apt-get install libssl-dev
 
       - name: Install cargo-hack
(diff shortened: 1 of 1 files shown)


jxs:
Hi, and thanks for this! Should we pin the version? And possibly configure dependabot for the repo CC @daniel-abramov

nickelc:
Pinning the version of `cargo hack`?

daniel-abramov (approved):

jxs:
> Pinning the version of `cargo hack`?

yeah, like it shows [here](https://github.com/taiki-e/install-action#example-workflow)

PR #263, opened 2023-02-15
Implement tungstenite::stream::NoDelay for MaybeTlsStream

zazabe:
`tungstenite::stream::NoDelay` is implemented for `tungstenite::MaybeTLsStream` since [v0.15.0](https://github.com/snapview/tungstenite-rs/commit/32450ae5af0070c7f3fd8341d244525119672506#diff-aa0a8d713ea64389a4b4916384d286b01c0de291edf10fec2f213123419e2647R128) but it's not implemented for the  `tokio_tungstenite::MaybeTLsStream` wrapper, this pull request fix this.

### DIFF
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -6,6 +6,7 @@
 use std::{
     pin::Pin,
     task::{Context, Poll},
+    io::{Read, Write},
 };
 
 use tokio::io::{AsyncRead, AsyncWrite, ReadBuf};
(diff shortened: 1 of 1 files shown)


daniel-abramov:
This is done for a reason. The `NoDelay` trait is not used inside `tokio-tungstenite`. There is a possibility to disable the Nagle algorithm though, see: https://github.com/snapview/tokio-tungstenite/pull/252

zazabe:
Ah ok, but this option to disable Nagle is not released yet, so my workaround currently is to rewrite `connect_async_with_config` in my crate and call [`std::net::TcpStream#set_nodelay()`](https://doc.rust-lang.org/std/net/struct.TcpStream.html#method.set_nodelay).

Wondering if it would not be more streamlined with `tungstenite` API if  `NoDelay` was implemented for `tokio_tungstenite::MaybeTLsStream` instead of introducing a `disable_nagle` parameter.

daniel-abramov:
Yeah, that's right, but we did not do it because according to https://github.com/snapview/tokio-tungstenite/pull/252#issuecomment-1383200591 we figured out that having the Nagle algorithm disabled by default (as we did in `tungstenite`) is not the best idea as disabling it brings more problems than solves, i.e. `tungstenite` had it implemented differently, but `tungstenite` always disabled the Nagle algorithm (which as we found out is not the best idea).

PR #270, opened 2023-03-17
[refactor] connect.rs: extract duplicated code into a function

vvv:
`connect_async_with_config` and `connect_async_tls_with_config` have identical implementations. `connector` parameter is the only difference.


daniel-abramov:
Thanks!

PR #274, opened 2023-04-13
Update dependencies and set rust-version

udoprog:
Supported rust-version before this change was 1.63, so upgrading dependencies had no effect.

This is particularly useful since tungstenite has upgrade to 0.19, but a few more dependencies are coming along for the ride as well.


daniel-abramov (approved):

daniel-abramov:
Is likely to be released tomorrow (currently the GitHub is [partially] down, so I can't really update the remote to update the CHANGELOG, etc).

Issue #141, opened 2020-12-23
Cargo.toml requirements unclear to run examples


ananswam:
I'm fairly new to Rust, so feel free to call me a noob and close this. I was trying to get the server example to run. I was able to do so easily by cloning and running `cargo run --example server 127.0.0.1:12345` as suggested.

However, when I started to incorporate that file content into my own project, I got all kinds of weird errors until I combed through the Cargo.toml file for tokio-tungstenite and added the following to my own Cargo.toml.

```
[dependencies]
tokio-tungstenite = "*"

futures-util = { version = "0.3", default-features = false, features = ["async-await", "sink", "std"] }
pin-project = "1.0"
futures-channel = "0.3"
tokio = { version = "0.3", default-features = false, features = ["io-std", "macros", "rt-multi-thread", "stream", "time"] }

[dependencies.tungstenite]
version = "0.11.1"
default-features = false
```

Initially, I tried just adding futures-util and futures-channel and tokio with the latest versions and default features based on what the IDE did not recognize, and I got compile errors. I think it must have been due to version incompatibility or feature selection.

In any case, I am all up and running now and excited to use the code, but it might be helpful to have the full above Cargo.toml snippet included with the example.

application-developer-DA:
Thanks for pointing out to that. Yeah, it's a bit boilerplate, which probably could be simplified at some point, so that we need less dependencies at the beginning. Tokio-related crates is something that we expect a person is familiar with since being not familiar with them makes the effective usage of `tokio-tunsgtenite` a bit complicated (as well as writing a new async-code).

Out of curiosity: why did you need to specify the `tunsgtenite` dependency? I may assume it's due to example using `tungstenite` explicitly, right? If so, I have a small hint: we re-expert the `tungstenite` in `tokio-tungstenite`, so that you don't need to include `tungstenite` dependency in your `Cargo.toml` directly to import the `Message` type, i.e. something like: `use tokio_tungstenite::tungstenite::Message` would do.

(Yes, I know, our example of server could be improved ;)

PR #275, opened 2023-04-16
Use re-exported tungstenite types so its easier to copy examples and …

alexgeek:
…not depend directly on tungstenite.

(I ran into the same confusion as in: https://github.com/snapview/tokio-tungstenite/issues/141)

### DIFF
--- a/examples/autobahn-server.rs
+++ b/examples/autobahn-server.rs
@@ -2,7 +2,10 @@ use futures_util::{SinkExt, StreamExt};
 use log::*;
 use std::net::SocketAddr;
 use tokio::net::{TcpListener, TcpStream};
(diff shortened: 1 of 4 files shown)


daniel-abramov:
Neat :) Would you mind checking the CI errors? It seems like the formatting is a bit off.

agalakhov:
JFYI: we already reexport Tungstenite as a whole:
```
pub use tungstenite;
```

alexgeek:
> JFYI: we already reexport Tungstenite as a whole:
> 
> ```
> pub use tungstenite;
> ```

Should I have done something differently in the PR because of that?

agalakhov:
No, everything looks fine. This only means that there is no need to depend directly of tungstenite even if using non-reexported types.

PR #281, opened 2023-05-05
Fix poll_flush on closed connection

CBenoit:
Flushing when there is nothing to send anymore should succeed.

Relevant code in `tungstenite` crate code:
https://github.com/snapview/tungstenite-rs/blob/e5efe537b87a6705467043fe44bb220ddf7c1ce8/src/protocol/mod.rs#L390


daniel-abramov (approved):
Looks sensible, thanks.

PR #287, opened 2023-06-17
`poll_ready` flush after `WouldBlock` errors

### LINKED ISSUE #286: Does #284 mean that the send buffer can be unbounded?
Hi there --

We've been looking at some of our uses of tokio-tungstenite and as part of that we came across #284. It looks like #284 makes it so that [`poll_ready` always returns `Poll::Ready`](https://github.com/snapview/tokio-tungstenite/blob/36b9d94ba23044bf96bca0f1e49cd37dd5dd4a3e/src/lib.rs#L324-L326).

However, consider what happens if we call `SinkExt::feed` repeatedly while never flushing (or if we try flushing in a `select!` loop but it always gets cancelled before it's done) -- it would seem that we keep adding messages to the tungstenite buffer and never flushing anything.

This seems to mean that there is no upper bound to the size of the buffer, i.e. that if we never call flush then the buffer can grow in an unbounded fashion.

1. Is my understanding of this code correct?
2. If so, then I think maybe we should consider changing this behavior. Rather than return to the pre-#284 status quo where we flush every time we call `poll_ready`, maybe flush if the buffer is larger than some size?

cc @alexheretic, @agalakhov

alexheretic:
With this PR `poll_ready` continues to generally return `Ready`. However, after `start_send` encounters `WouldBlock` `poll_ready` will instead try to flush.

When poll_ready->flush returns `Ready` poll_ready will revert to just returning `Ready` ... until the next `WouldBlock`.

This mechanism prevents the case tokio-tungstenite users filling up the tungstenite write-buffer without actually handling any errors (since `WouldBlock` errors are not returned to the user).

We also maintains `feed` generally _not_ flushing as it will only do so after encountering `WouldBlock` errors. So should perform well.

See https://github.com/snapview/tokio-tungstenite/issues/286#issuecomment-1595678753.

Resolves #286 

## Questions/alternatives
* We could do a similar thing by checking if the write-buffer is fuller than `WebSocketConfig::write_buffer_size`. I think this would achieve the same thing though it would require some changes upstream to expose the current write-buffer size. I'm not sure it's actually better than just tracking would-block as this PR does though.
* I initially just had `poll_ready` simply return `Pending` after `WouldBlock` but trying to flush in this case seems more correct & useful, plus seems to fit with the `Sink` docs
    > Implementations of `poll_ready` and `start_send` will usually involve flushing behind the scenes in order to make room for new messages. It is only necessary to call `poll_flush` if you need to guarantee that *all* of the items placed into the `Sink` have been sent.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -326,11 +326,15 @@ where
 {
     type Error = WsError;
 
-    fn poll_ready(self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+    fn poll_ready(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
         if self.ready {
             Poll::Ready(Ok(()))
         } else {
-            Poll::Pending
+            // Currently blocked so try to flush the blockage away
+            (*self).with_context(Some((ContextWaker::Write, cx)), |s| cvt(s.flush())).map(|r| {
+                self.ready = true;
+                r
+            })
         }
     }
 
@@ -355,10 +359,10 @@ where
     }
 
     fn poll_flush(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
-        self.ready = true;
         (*self).with_context(Some((ContextWaker::Write, cx)), |s| cvt(s.flush())).map(|r| {
-            // WebSocket connection has just been closed. Flushing completed, not an error.
+            self.ready = true;
             match r {
+                // WebSocket connection has just been closed. Flushing completed, not an error.
                 Err(WsError::ConnectionClosed) => Ok(()),
                 other => other,
             }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -193,6 +193,11 @@ pub struct WebSocketStream<S> {
     inner: WebSocket<AllowStd<S>>,
     closing: bool,
     ended: bool,
+    /// Tungstenite is probably ready to receive more data.
+    ///
+    /// `false` once start_send hits `WouldBlock` errors.
+    /// `true` initially and after `flush`ing.
+    ready: bool,
 }
 
 impl<S> WebSocketStream<S> {
@@ -226,7 +231,7 @@ impl<S> WebSocketStream<S> {
     }
 
     pub(crate) fn new(ws: WebSocket<AllowStd<S>>) -> Self {
-        WebSocketStream { inner: ws, closing: false, ended: false }
+        Self { inner: ws, closing: false, ended: false, ready: true }
     }
 
     fn with_context<F, R>(&mut self, ctx: Option<(ContextWaker, &mut Context<'_>)>, f: F) -> R
@@ -322,18 +327,27 @@ where
     type Error = WsError;
 
     fn poll_ready(self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
-        Poll::Ready(Ok(()))
+        if self.ready {
+            Poll::Ready(Ok(()))
+        } else {
+            Poll::Pending
+        }
     }
 
     fn start_send(mut self: Pin<&mut Self>, item: Message) -> Result<(), Self::Error> {
         match (*self).with_context(None, |s| s.write(item)) {
-            Ok(()) => Ok(()),
+            Ok(()) => {
+                self.ready = true;
+                Ok(())
+            }
             Err(WsError::Io(err)) if err.kind() == std::io::ErrorKind::WouldBlock => {
-                // the message was accepted and queued
-                // isn't an error.
+                // the message was accepted and queued so not an error
+                // but `poll_ready` will start returning pending now.
+                self.ready = false;
                 Ok(())
             }
             Err(e) => {
+                self.ready = true;
                 debug!("websocket start_send error: {}", e);
                 Err(e)
             }
@@ -341,6 +355,7 @@ where
     }
 
     fn poll_flush(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.ready = true;
         (*self).with_context(Some((ContextWaker::Write, cx)), |s| cvt(s.flush())).map(|r| {
             // WebSocket connection has just been closed. Flushing completed, not an error.
             match r {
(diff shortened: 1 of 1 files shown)


alexheretic:
By writing the question _"Should poll_ready try to flush if it is !ready?"_ I've mostly convinced myself that it should.

Someone stop me or I'll add that to this PR in a bit...

daniel-abramov:
It looks like a quick fix, but it feels a bit risky as we need to make many assumptions about how `tungstenite` should buffer the messages. I'm also not sure if there are no edge cases that we need to consider, e.g. the latest changes in `tungstenite` allow specifying how much data would be buffered and **not written** to the underlying stream before hitting a limit, which theoretically means that if a size of that limit/threshold being close to the size of our buffer, we can get to the point where N `feed()`s (i.e. `poll_ready()` + `start_send()`) called in a row would simply buffer N messages filling up the internal out buffer in `tungstenite` without writing anything to the stream causing the N+1 attempt to call `feed()` to return an error telling that the buffer is full.

I don't think there is **an elegant** way to fix this issue (surely we could add some `if`s comparing the threshold value ad the out buffer with the frame size, but I'm not sure if it's really worth the effort and if it introduces too many complexities).

[Personally, I think](https://github.com/snapview/tokio-tungstenite/issues/286#issuecomment-1595798441) that if we really want to have the maximum flexibility for the buffering/flushing, it would be better to introduce explicit methods for that and keep the former implementation of the `Sink` (that seem to have always behaved in a correct [albeit inefficient] way).

alexheretic:
I think it is important that `feed` in general does not flush. This is the reason `feed` and `send` exist. Having both of them just always flush seems just incorrect and a performance bug. We don't need to do that anymore. It seems a bit silly to have `feed` behave incorrectly in order to ensure it behaves ... correctly.

To be clear the reason for `feed` to exist is:

> Unlike send, the returned future does not flush the sink.

I don't see the configurable `write_buffer_size` & `max_write_buffer_size` as an edge case, it is just a buffering behaviour within the general contract of a write interface. If the user sets the `max_write_buffer_size` low enough they may see an error. A simpler example, if a user sets the `max_write_buffer_size` smaller than the size of a message all `feed` attempts will fail no matter how you flush. Can you explain why this breaks the `Sink` contract?

As I see it now this behaviour looks correct. `feed` generally won't flush, but now _will_ try to `flush` when writes block. This matches the `Sink` docs suggestions for behaviour.

> Implementations of poll_ready and start_send will usually involve flushing behind the scenes in order to make room for new messages. It is only necessary to call poll_flush if you need to guarantee that all of the items placed into the Sink have been sent.

All other tungstenite errors are simply returned to the user to deal with. That isn't a change from existing behaviour. So the only interesting error is `WouldBlock` since it is _not_ returned to the user. And that was the issue with https://github.com/snapview/tokio-tungstenite/pull/284 that it didn't handle this hidden error any differently from the flush happy past.

This PR addresses that by flushing after `WouldBlock` errors. That means in general `feed` will not flush (which is what it exists to not do) but if the underlying stream starts blocking it _will_ try to flush meaning we'll either flush or return `Pending` and reflect the blocking properly. I still think this PR seems a good fix for the issue.

jgallagher:
> I don't see the configurable write_buffer_size & max_write_buffer_size as an edge case, it is just a buffering behaviour within the general contract of a write interface. If the user sets the max_write_buffer_size low enough they may see an error. A simpler example, if a user sets the max_write_buffer_size smaller than the size of a message all feed attempts will fail no matter how you flush. Can you explain why this breaks the Sink contract?

If I set `max_write_buffer_size` to a value that is in between the size of my messages and the size at which we start seeing `WouldBlock` errors, what happens if I call `feed()` repeatedly without flushing? I think (just from reading the changes, not trying it, so could certainly be wrong!) that:

1. We would never set `self.ready` (because we haven't sent enough data to start seeing `WouldBlock`), so `poll_ready()` always returns `Ready`
2. Once we hit `max_write_buffer_size`, we'd get a failure in `start_send`

This seems to break the contract of `feed`, right? I would expect to be able to `feed()` continuously and the `Sink` should flush as needed, which wouldn't happen in this case. Maybe my configuration is _questionable_, but it seems like it should be valid? I've based my `max_write_buffer_size` on the size of my messages (and ensure it's large enough to hold some number of them), but because it's smaller than some underlying value I don't know I hit this edge case.

alexheretic:
Firstly thinking about hypotheticals, `max_write_buffer_size` should never be set below `write_buffer_size` and ideally its always at least `write_buffer_size + 1 message`. We should probably add some basic errors around this upstream. But in the following I'll assume these two are set to reasonable things or default.

> We would never set self.ready (because we haven't sent enough data to start seeing WouldBlock), so poll_ready() always returns Ready

If you run your example you'll see this isn't the case. It's because the write-buffer _isn't_ filling up. If the underlying stream is not returning `WouldBlock` (or any other error that you would see) that means the writes are going into the underlying stream and not building up in tungstenite's write-buffer.

> I've based my max_write_buffer_size on the size of my messages (and ensure it's large enough to hold some number of them)

As mentioned before it should always be at least `write_buffer_size + 1 message`. I don't really see the benefit of setting bounding it unless you aim to ignore most errors and only handle `WriteBufferFull`. 

The important point is the **write-buffer only fills up past `write_buffer_size` when the underlying stream writes are failing**. 

I think I have clearly not documented it well enough upstream :( Here's what I have currently, let me know if you have suggestions!

> write_buffer_size: [usize](https://doc.rust-lang.org/1.70.0/std/primitive.usize.html)
> 
> The target minimum size of the write buffer to reach before writing the data to the underlying stream. The default value is 128 KiB.
> 
> Note: [flush](file:///home/alex/.cache/cargo/target/doc/tungstenite/protocol/struct.WebSocket.html#method.flush) will always fully write the buffer regardless.
> max_write_buffer_size: [usize](https://doc.rust-lang.org/1.70.0/std/primitive.usize.html)
> 
> The max size of the write buffer in bytes. Setting this can provide backpressure in the case the write buffer is filling up due to write errors. The default value is unlimited.
> 
> Note: Should always be set higher than [write_buffer_size](file:///home/alex/.cache/cargo/target/doc/tungstenite/protocol/struct.WebSocketConfig.html#structfield.write_buffer_size).

daniel-abramov:
> imo the approach in this pr seems good, I've tried to explain why as clearly i can. Are there remaining concerns @daniel-abramov?

Apologies for the long delay.

Let's give it a try, thanks for explaining your point!

daniel-abramov (approved):

PR #298, opened 2023-08-22
update rustls to fix RUSTSEC-2023-0053

lightydo:


JamesHinshelwood:
Hi @daniel-abramov @alexheretic , sorry for the ping. Would it be possible to merge this soon and release a new version, so upstream projects can resolve the vulnerability alert?

src/tls.rs
```diff
@@ -104,8 +104,8 @@ mod encryption {
                             }
                             #[cfg(feature = "rustls-tls-webpki-roots")]
                             {
-                                root_store.add_server_trust_anchors(
-                                    webpki_roots::TLS_SERVER_ROOTS.0.iter().map(|ta| {
+                                root_store.add_trust_anchors(
```
sdroege, line 107: The updated version requirements in Cargo.toml are needed because this function does not exist in older versions? If so, only the version of rustls has to be bumped or not? Why also the bump of tokio-rustls?

Cargo.toml
```diff
@@ -55,11 +55,11 @@ version = "0.3.1"
 
 [dependencies.tokio-rustls]
 optional = true
-version = "0.24.0"
+version = "0.24.1"
 
 [dependencies.webpki-roots]
 optional = true
-version = "0.23.0"
+version = "0.25.2"
```
sdroege, line 62: IIRC the webpki-roots API is exported somewhere from tokio-tungstenite, so bumping this would require releasing this as a breaking release (i.e. 0.21.0).

alexheretic: I don't believe it is exported. I can only see [the one usage](https://github.com/snapview/tokio-tungstenite/blob/71d06d4495b6315181ced85b82a28044d3383854/src/tls.rs#L108) which is afaics isolated from the public api.

src/tls.rs
alexheretic: Yes looks like `add_trust_anchors` was added in _rustls_ `0.21.6` and `add_server_trust_anchors` deprecated.

I don't know what changed in _tokio-rustls_ `0.24.1`. But it's probably harmless to up the min version anyway :shrug:

alexheretic (approved):
lgtm

src/tls.rs
sdroege: If there's no reason then it forces users to update for no reason, and that might actually break something for them if for whatever reason they must use an older version. I wouldn't unnecessarily bump minimum versions.

Cargo.toml
sdroege: I misremembered then, sorry :) Thanks for checking

src/tls.rs
ZoeS17: It looks like a refactor happened around commit `07e8da6e52f0eb76323f77d441bd1a06498b2f1c` [here](https://github.com/rustls/tokio-rustls/compare/main%40%7B07e8da6e52f0eb76323f77d441bd1a06498b2f1c%7D...main)'s the diff from `0.24.0` to `0.24.1` for tokio-rustls as I could find it.

daniel-abramov (approved):

agalakhov:
It's time to release new version.

PR #306, opened 2023-11-10
chore(rustls): switch to new rustls versions

kskalski:
New `*-rustls` versions introduce some breaking API changes, but they do not seem to impact `tokio-tungstenite` directly, instead the hard constraint on minor version of them prevents from using `tokio-tungstenite` with them.
Alternatively, I think loosening the dependencies to any minor versions (e.g. for `rustls` using  `version = "0.21"`, which allows cargo to dump automatically to `0.22.*` could work too).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,7 @@ version = "0.2.11"
 
 [dependencies.rustls]
 optional = true
-version = "0.21.6"
+version = "0.22.0-alpha.3"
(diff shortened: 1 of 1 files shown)


daniel-abramov:
Thanks for the information. Note, that these are alpha versions (not something we would want to depend on), but switching to the stable versions once they are released would b certainly fine!

PR #316, opened 2023-12-26
Simplify response body type in `server-custom-accept` example

nickelc:
The empty `Full<T>::default()` body behaves like the `Empty` implementation.


daniel-abramov (approved):
Cool, I even forgot we used that thing 🙂

Issue #334, opened 2024-06-02
tokio-tungstenite not using version 0.22 of tungstenite


Its-Just-Nans:
Hi @daniel-abramov 

Looks like the tungstenite https://github.com/snapview/tokio-tungstenite/blob/c1025d59ff15eeefc2dc07495fd6a8365cc6e0af/Cargo.toml#L36 is not at 0.22

That leads to have two **differents** versions of tungstenite in the same project

Is that normal ?

Maybe a solution would be to bump the two crates to 0.23
- `tungstenite@0.22` **-> `tungstenite@0.23`**
- `tokio-tungstenite@0.22` using `tungstenite@0.21` **-> `tokio-tungstenite@0.23` using `tungstenite@0.23`**

(and it will fix  https://github.com/snapview/tungstenite-rs/pull/427#pullrequestreview-2092151225)

daniel-abramov:
Thanks for reporting that. I'm going to address that in a minute!

PR #335, opened 2024-06-02
prepare 0.23 ? #334

Its-Just-Nans:
See https://github.com/snapview/tokio-tungstenite/issues/334

This PR needs tungstenite 0.23

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,52 +1,58 @@
-# 0.22.0
+# CHANGELOG
+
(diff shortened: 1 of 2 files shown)


daniel-abramov (commented):
The diff contains a bit more (unintended) changes such as changing the headers in the CHANGELOG for previous releases (probably by accident?).

Also, a small change is required still. Nevertheless, thanks for submitting! I'm going to fix that and publish a release soon.

Issue #92, opened 2020-03-23
send headers along with `connect_async`


EverlastingBugstopper:
I need to send along some authentication headers along with a request to instantiate a websocket connection, is there a way I can do this?

application-developer-DA:
Sure, as you can [see in the docs](https://docs.rs/tokio-tungstenite/0.10.1/tokio_tungstenite/fn.connect_async.html), the `connect_async` function takes a generic request (any type which implements a trait`. This trait is implemented for `httparse::Request`, which [allows you](https://docs.rs/httparse/1.3.4/httparse/struct.Request.html) to construct request, containing the required headers.

EverlastingBugstopper:
oh, nice! thanks :)

elderhammer:
My solution is to use [http::Request](https://docs.rs/http/0.2.9/http/request/struct.Request.html).
```Rust
let request = Request::builder()
    .uri(address)
    .header(auth_key, auth_value)
    .header("sec-websocket-key", "foo")
    .header("host", "server.example.com")
    .header("upgrade", "websocket")
    .header("connection", "upgrade")
    .header("sec-websocket-version", 13)
    .body(())
    .unwrap();
```

PR #342, opened 2024-08-03
Additional Documentation of `IntoClientRequest` on `connect_async`

Brendan-Blanchard:
I came across https://github.com/snapview/tokio-tungstenite/issues/92 when I needed this behavior for a use case of mine and wanted to save future searchers the trouble with an explicit example in the docs. 

Let me know if there's anything I can change. My IDE didn't seem pleased with the formatting, even when adding the hidden `async fn example() { ... }` around it.

### DIFF
--- a/src/connect.rs
+++ b/src/connect.rs
@@ -16,23 +16,17 @@ use crate::{domain, stream::MaybeTlsStream, Connector, IntoClientRequest, WebSoc
 /// complex uses.
 ///
 /// ```no_run
-/// # async fn example() {
+/// # use tungstenite::client::IntoClientRequest;
+///
+/// # async fn main() {
 /// use tungstenite::http::{Method, Request};
 /// use tokio_tungstenite::connect_async;
 ///
-/// let request = Request::builder()
-///             .uri("wss://api.example.com")
-///             .method(Method::GET)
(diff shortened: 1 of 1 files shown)


src/connect.rs
```diff
@@ -10,6 +10,30 @@ use tungstenite::{
...
+/// let request = Request::builder()
+///             .uri("wss://api.example.com")
+///             .method(Method::GET)
+///             .header("Sec-WebSocket-Key", "someUniqueValue")
+///             .header("Sec-WebSocket-Version", "13")
+///             .header("host", "api.example.com")
+///             .header("Connection", "Upgrade")
+///             .header("Upgrade", "websocket")
+///             .body(())
+///             .unwrap();
```
daniel-abramov, line 32: I think this kind of pattern is actually a bit problematic as it requires specifying all WebSocket headers properly. A more typical issue that people experience is that they just want to add one or more additional headers on top of what we generate automatically, see https://github.com/snapview/tungstenite-rs/issues/327

Brendan-Blanchard: Ah, that's even better for my use case, too. I've updated the example if you're okay with adding it. Thanks!

PR #348, opened 2024-09-04
deps: update `rustls-native-certs` to 0.8

nickelc:
The `load_native_certs()` function now returns all errors instead of raising only the first error.

Not finding any native root CA certificates is not fatal if the "rustls-tls-webpki-roots" feature is enabled.

### DIFF
--- a/src/tls.rs
+++ b/src/tls.rs
@@ -95,10 +95,26 @@ mod encryption {
                             let mut root_store = RootCertStore::empty();
                             #[cfg(feature = "rustls-tls-native-roots")]
                             {
-                                let native_certs = rustls_native_certs::load_native_certs()?;
-                                let total_number = native_certs.len();
+                                let rustls_native_certs::CertificateResult {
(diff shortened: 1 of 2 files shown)


src/tls.rs
```diff
@@ -95,10 +95,26 @@ mod encryption {
...
+
+                                if !errors.is_empty() {
+                                    log::warn!(
+                                        "native root CA certificate loading errors: {errors:?}"
+                                    );
+                                }
+
+                                // Not finding any native root CA certificates is not fatal if the
+                                // "rustls-tls-native-roots" feature is enabled.
+                                #[cfg(not(feature = "rustls-tls-native-roots"))]
```
sdroege, line 110: This is in a `#[cfg(feature = "rustls-tls-native-roots")]` block, so would never be compiled in

nickelc: Thanks, it should be "rustls-tls-webpki-roots".

PR #358, opened 2025-01-13
remove Unpin requirement for the inner stream

BiagioFesta:
Currently, `WebSocketStream` [requires][1] `Unpin` being implemented for the underlying I/O resource.

The aim of this PR is to remove this constrain.

### Context
Any underlying I/O streams that work with timeouts logic *won't* implement `Unpin`.

For example: [`tokio-io-timeout::TimeoutStream`][2] it does **not** implement `Unpin`.

That's because any timer related future over tokio runtime is `!Unpin`.

Of course, an easy solution would be to allocate the underlying stream on the heap for pinning (e.g., `Pin<Box<T>>`). However, that means heap allocation and pointer indirection. Although those drawbacks are probably not really impactful, removing `Unpin` requirement will make this library more flexible.

On the other hand, removing `Unpin` does not come for free:
1. Adding additional dependency: `pin_project`. 
   * I think this is not a real concern. `pin_project` is very **very** widely spread. It is likely indirectly pulled in anyway.
2. `unsafe` code.
   * This is real drawback of this PR. The tread-off here is: library more flexible but with `unsafe` blocks. 

[1]: https://docs.rs/tokio-tungstenite/0.26.1/tokio_tungstenite/struct.WebSocketStream.html#impl-Stream-for-WebSocketStream%3CT%3E
[2]: https://docs.rs/tokio-io-timeout/latest/tokio_io_timeout/struct.TimeoutStream.html#impl-Unpin-for-TimeoutStream%3CS%3E

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -189,7 +189,9 @@ where
 /// them in `futures-rs` crate documentation or have a look on the examples
 /// and unit tests for this crate.
 #[derive(Debug)]
+#[pin_project::pin_project]
 pub struct WebSocketStream<S> {
+    #[pin]
     inner: WebSocket<AllowStd<S>>,
     closing: bool,
     ended: bool,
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -330,59 +347,98 @@ where
...
-            }
-        })
+        self.as_mut()
+            .with_context(Some((ContextWaker::Write, cx)), |s| unsafe {
+                // SAFETY: library's `Write` impl is going to Pin anyway
+                cvt(s.get_unchecked_mut().flush())
+            })
+            .map(|r| {
+                unsafe {
+                    self.get_unchecked_mut().ready = true;
```
BiagioFesta, line 405: missing SAFETY comment:
```
// SAFETY: not moving out
```

Dushistov:
>  think this is not a real concern. pin_project is very very widely spread

tokio depend on pin-project-lite, not pin-project. futures-util also depends on pin-project-lite.

BiagioFesta:
> > think this is not a real concern. pin_project is very very widely spread
> 
> tokio depend on pin-project-lite, not pin-project. futures-util also depends on pin-project-lite.

I guess we can switch to `pin-project-lite`, still not the real concern of this change I guess

PR #367, opened 2025-09-11
Implement into_inner to get the underlying stream

hi-abhi:
Followup to the analogous change in tungstenite (https://github.com/snapview/tungstenite-rs/pull/516) to get the underlying raw stream from a WebSocketStream.

There's an argument that into_inner on the `WebSocketStream` should return the `WebSocket` and then consumers call into_inner again, but that wraps AllowStd which is not public so I decide to return the underlying stream only which I think is fine.

This is a bit of an odd pull request in that it doesn't compile because a new version of tungstenite hasn't been cut yet. Once `tungstenite = 0.28` is cut, hopefully this can be merged and be part of `tokio-tungstenite = 0.28`

### DIFF
--- a/src/compat.rs
+++ b/src/compat.rs
@@ -96,6 +96,11 @@ impl<S> AllowStd<S> {
             }
         }
     }
+
+    /// Returns the underlying stream.
+    pub(crate) fn into_inner(self) -> S {
+        self.inner
+    }
 }
 
 // Proxy Waker that we pass to the internal AsyncRead/Write of the
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -247,6 +247,11 @@ impl<S> WebSocketStream<S> {
         f(&mut self.inner)
     }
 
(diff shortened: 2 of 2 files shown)


hi-abhi:
Checking back in on this, what type of release process/schedule do you have for tungstenite-rs and this crate? There's been a number of commits to tungstenite since 0.27 was cut, hoping this can make it in for 0.28 if it's not much trouble.

Thanks :)

daniel-abramov:
> what type of release process/schedule do you have for tungstenite-rs and this crate

We typically release a new version when there are substantial changes to the library, such as a noticeable performance improvement, a bug fix, a new feature, or a change in the behavior of an existing feature. There was no release since there have been no functional or performance changes (apart from some tiny ones) in `tungstenite` since `0.27.0`.

> hoping this can make it in for 0.28 if it's not much trouble

Alright, I've just published `0.28.0`. Feel free to update the PR :)

